PR #10, opened 2025-11-11
fix multi-call solver

anthonyduong9:
## This PR contains:
- [ ] New features
- [ ] Changes to dev-tools e.g. CI config / github tooling
- [ ] Docs
- [x] Bug fixes
- [ ] Code refactor

### What is the current behavior? (You can also link to an open issue here)

If a solver runs an agent multiple times (by iterating over messages, and for each message, appending to the agent state's messages and running the agent), then each time, the agent joins all of the user messages and passes this as the prompt to the command.

### What is the new behavior?

Each time, the agent checks if there are user messages after which there have been no assistant messages and if there are, only joins those, and resumes previous conversation.

### Does this PR introduce a breaking change? (What changes might users need to make in their application due to this PR?)

No.

### Other information:

Users have been working around the current behavior for Codex CLI by using just one sandbox agent bridge and `codex resume --last`.

### DIFF
--- a/src/inspect_swe/_claude_code/claude_code.py
+++ b/src/inspect_swe/_claude_code/claude_code.py
@@ -10,12 +10,7 @@ from inspect_ai.agent import (
     agent_with,
     sandbox_agent_bridge,
 )
-from inspect_ai.model import (
-    ChatMessageAssistant,
-    ChatMessageSystem,
-    ChatMessageUser,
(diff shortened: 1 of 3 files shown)


anthonyduong9:
> Thank you! Looks great, just a couple of comments about factoring out common code and establishing a constraint on the input.

No problem, thanks!

PR #19, opened 2026-01-13
WIP gemini-cli

alexlloyd-apollo:
This is currently a quite hacky, mostly vibe-coded implementation, built just to get something that works.

TODO:
1. check gemini's cli params for what we want to support
2. We currently download a fixed node version to the sandbox as gemini-cli doesn't have a standalone binary. I couldn't find a good way to avoid this but maybe there is one.


alexlloyd-apollo:
Closed in favor of #22

PR #32, opened 2026-03-05
add pip dependency

aokellermann:
The following code uses `pip` during runtime, but it is currently not declared as a dependency:

https://github.com/meridianlabs-ai/inspect_swe/blob/e81b4764d0a7a960d87b51984ee7b64c280fd8a8/src/inspect_swe/_util/agentwheel.py#L344-L358


jjallaire:
Alternate resolution with https://github.com/meridianlabs-ai/inspect_swe/pull/46

PR #42, opened 2026-04-21
fix(solvers): forward user= to sandbox_agent_bridge

art-dsit:
user is missing being passed through in a few places


art-dsit:
Relies on https://github.com/UKGovernmentBEIS/inspect_ai/pull/3595

PR #46, opened 2026-04-30
Bootstrap pip for wheel downloads

Scott-Simmons:
## Summary

https://github.com/UKGovernmentBEIS/inspect_evals/pull/1496 found that pip not being available would raise an error.

- bootstrap pip via ensurepip when missing before running pip download
- include ensurepip failure details in the raised error


Scott-Simmons:
I see https://github.com/meridianlabs-ai/inspect_swe/pull/32 is also raised. Can close out whichever works for the project

jjallaire (approved):

PR #53, opened 2026-05-17
Fix opencode native skills

tkukurin:
Failure mode: opencode tried to download rg from GitHub at tool-call time and failed with a transport error

quickfix in this PR: install rg not just node. OpenCode discovers global skills under `$HOME/.config/opencode/skills` and uses its native skill tool to load.

I'm not 100% sure this is the best way; to demo failure mode + fix, try eg `uv run inspect eval examples/skills/task.py@agent_skills --model openai/gpt-4.1-mini -T agent=opencode`

ref:
https://github.com/anomalyco/opencode/blob/be6a89a3b89bcdbd359948078360591a84e91f04/packages/opencode/src/tool/skill.ts#L42


jjallaire:
Could you run make check and resolve the ruff issues.

PR #63, opened 2026-06-03
claude_code: report real served model instead of `inspect` sentinel

alexlloyd-apollo:
## Motivation

`inspect_swe` bridges the inner agent's model calls to the Inspect task model by passing a sentinel model string (`"inspect"` / `"inspect/<model>"`) to the inner binary. For `claude_code` that sentinel was also the model's *displayed* identity, so its system prompt read:

> - You are powered by the model **inspect**.

This is an eval-awareness / realism artifact.

## Change

- `claude_code` now presents the **real served model name** (bare `Model.name`, e.g. `claude-sonnet-4-5`) on `--model` and the `ANTHROPIC_*` env vars. New **`presented_model`** param overrides it (defaults to the served model); it's purely the displayed identity — calls are still bridged to the served model.
- `model_aliases` is keyed on the presented name → the real `Model`, with the `"inspect"`/`"inspect/<model>"` sentinel kept as the bridge `model=` **fallback**, so routing is unchanged. Caller-supplied `model_aliases` still takes precedence (also fixes an accidental shadowing that dropped it).
- The `opus/sonnet/haiku/subagent_model` env vars now carry real names instead of the sentinel; a *set* sub-model registers its own alias so it routes to its own model (the fallback previously collapsed them onto the main model). Unset roles inherit the primary name (no-op vs. before).
- Removed the now-unused `_util/model.py` helper.

**Scope:** `claude_code` only — `codex_cli` / `gemini_cli` / `opencode` / `mini_swe_agent` don't surface the `inspect` sentinel and are unchanged.

## Testing

- New `tests/test_presented_model_live.py` (Docker + provider): asserts the inner system prompt reports the real model (not `inspect`) and routing still resolves to the served model, for native (`anthropic/claude-sonnet-4-5`) and cross-provider (`openai/gpt-5`). Both pass.
- `make check` clean; existing `test_multi_call.py` (all agents) unchanged.


jjallaire:
Thanks for this @alexlloyd-apollo — clean fix for the eval-awareness artifact. I pushed a follow-up commit on top of yours (left your commit intact for credit):

- **Renamed `presented_model` → `model_config`** to match the equivalent knob in `codex_cli` (where `--model` is likewise cosmetic and the real model is bridged), keeping the two agents consistent.
- **Verified the genuine-identity behavior against the Claude Code binary** (2.1.160) and fixed the live test. The identity block is built from an internal catalog: `hX()` canonicalizes the id via substring match (dated / `[1m]` / provider-prefixed forms) and `JD()`/`I4q()` render the friendly name + cutoff. So a recognized Anthropic id (clean *or* dated) already yields the genuine *"…powered by the model named Sonnet 4.5. The exact model ID is …"* block — no normalization needed on our side — and an unrecognized id (cross-provider, or a Claude model newer than the installed CC) honestly degrades with no fabricated Claude identity.
- The previous test asserted `"powered by the model claude-sonnet-4-5"`, which only matches the **degraded** phrasing, so it was not verifying genuineness. Updated it to assert the genuine block, plus cases for a dated id, cross-provider honest-degrade, and a `model_config` override.

Routing, the per-role alias machinery, and the `_util/model.py` removal are unchanged from your commit. Happy to tweak the naming if you would prefer.

PR #65, opened 2026-06-05
claude_code: per-execution isolation for session, port, and config

alex-remedios:
## Summary

Three small fixes in `claude_code.py` that all move per-execution state out
of agent-instance scope, so the bridge behaves correctly under concurrent
samples, multi-epoch runs, and shared hosts. Each commit is independent;
none change the single-sample default-environment behavior.

### 1. Allocate session_id per `execute()` call (`51b0142`)
Previously created once at agent-instance build time. Multiple samples
sharing an agent instance would collide on a single Claude Code session id,
hitting "session already in use". Moved the `uuid.uuid4()` inside `execute`.

### 2. Honor `HOME` from `agent_env` in `_seed_claude_config` (`f3d4dda`)
`_seed_claude_config` writes to `$HOME/.claude/settings.json` via the
sandbox shell. Without forwarding the caller's `agent_env`, `sandbox="local"`
runs would write to the *real* `~/.claude/settings.json` of the host user
instead of the redirected `HOME=/tmp/...` that callers pass for test
isolation. Added `agent_env` param and forward it as `env=` to the
`sbox.exec` call.

### 3. Use an OS-assigned free port for the agent bridge (`4f2a776`)
`port = store().get(MODEL_PORT, 3000) + 1` made every run's first sample
try to bind `127.0.0.1:3001`. On shared hosts where another process holds
3001 (multi-user GPU clusters, dev machines with stray services), this
fails with "address already in use" before the first sample can run. Bind
to port 0 and read back the OS-assigned port; brief TOCTOU window between
probe-close and bridge-bind, but no worse than the previous scheme and
removes the cross-user collision class.

## Test plan

- [x] Live-mode end-to-end eval against the new bridge port code completed
      successfully (downstream consumer of `inspect_swe`).
- [ ] No new tests added — these are behavior-preserving for the default
      single-sample / non-shared-host path, which the existing test suite
      exercises.

### DIFF
--- a/src/inspect_swe/_claude_code/_events/live_consumer.py
+++ b/src/inspect_swe/_claude_code/_events/live_consumer.py
@@ -64,6 +64,7 @@ from inspect_ai.model._chat_message import (
     ChatMessageUser,
 )
 from inspect_ai.model._model import ModelEventSink
+from inspect_ai.model._model_output import StopReason
 from inspect_ai.util._span import current_span_id
 
 from .toolview import tool_view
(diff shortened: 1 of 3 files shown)


jjallaire:
Thanks for this! Ping me when this is out of draft and will review.

PR #69, opened 2026-06-15
feat(acp,codex,claude_code): session resume

jannes-apollo:
## Summary

Adds session resume to ACP agents: re-attach to an existing session, or resume from a prior conversation you supply (real, synthetic, or truncated). Useful for:
- branching and resuming conversations at a specific turn
- resampling from a chosen turn
- continuation evals — does the agent resume malign trajectories?

Prior turns become real session history for the CLI.

**Drafted by Opus 5:** 

## The API

`ChatMessage` is the front door, and it's generic across agents:

```python
# resume from a prior conversation — codex or claude code
agent = interactive_claude_code(resume_messages=prior)
agent = interactive_codex_cli(resume_messages=prior)

# re-attach to a session the CLI already persisted (any agent advertising loadSession)
agent = interactive_gemini_cli(resume_session_id=session_id)

# branch: resume claude code only up to one transcript row
spec = build_transcript(cwd="/root", items=[...], model="claude-opus-5")
agent = interactive_claude_code(resume_transcript=spec, resume_message_uuid=spec.item_uuids[1])
```

**`acp/agent.py`** — `resume_session_id` and `resume_messages` on `ACPAgentParams`. Subclasses materialize the on-disk session by overriding `_resolve_resume_session`; the base class verifies the server advertised `loadSession` before anything is written, and raises a clear `ACPError` for an agent that can't serialize a synthetic prior. `_load_session_meta` lets a subclass reach CLI-specific resume options through the request's `_meta`.

**Codex (`acp/_agents/codex_cli/rollout.py`)** — `build_rollout(prior=...)` takes `Sequence[PriorItem | ChatMessage]` and writes codex's `rollout-*.jsonl` into `$CODEX_HOME/sessions/...`; `parse_rollout` reads a real or saved rollout back into typed items plus session metadata, so you can truncate at a node and rebuild. `prior_from_messages` / `messages_from_prior` / `ParsedRollout.as_messages` convert both ways.

**Claude Code (`acp/_agents/claude_code/transcript.py`)** — `build_transcript` / `parse_transcript` for the on-disk format (`$CLAUDE_CONFIG_DIR/projects/<cwd-slug>/<session-id>.jsonl`: one row per *content block*, chained by `parentUuid`, `message` holding a raw Anthropic message). `resume_message_uuid` truncation goes through the Agent SDK's `resumeSessionAt`, so branching needs no transcript rewrite. New `config_dir` (`CLAUDE_CONFIG_DIR`), the counterpart of codex's `home_dir`.

## Why the typed items exist alongside ChatMessage

Per review feedback, `ChatMessage` is the documented default. The CLI-native item types stay as the full-fidelity layer, because `ChatMessage` can't express:

- **`custom_tool_call`** (codex's `apply_patch` path) takes free-form text, not JSON arguments; `ToolCall.arguments` is a dict.
- **Reasoning plaintext.** Inspect's Responses conversion emits reasoning with `content=[]` — correct for an API request, wrong for on-disk history, and the resumed CoT is exactly what a continuation eval is studying. The mapping here keeps plaintext, summary and ciphertext, and round-trips all three.
- **Rows with no Inspect equivalent** (`web_search_call`, `tool_search_call`) and session-level fields (`base_instructions`, `cli_version`, `turn_context`), which `parse_rollout` needs to preserve so a real session survives truncate-and-rebuild.

The conversion is hand-rolled rather than calling `inspect_ai.model._openai_responses.openai_responses_inputs`, because that module imports the `openai` SDK — not a runtime dependency of inspect_swe. Happy to swap to it, or to a provider-free version exported from `inspect_ai`, if you'd prefer either.

## Gotchas worth knowing

- **Thinking blocks are dropped from message-derived Claude Code priors by default.** Anthropic validates thinking signatures, so replaying one that didn't come from the same provider fails the next turn (`include_thinking=True` to keep them).
- **The Agent SDK resolves symlinks before slugging the cwd**, so the transcript is placed under the sandbox-resolved path. Writing under the unresolved spelling lands where `session/load` never looks — and resume then *silently* starts a fresh conversation rather than erroring.
- **A synthetic prior is readable by the agent.** `CODEX_HOME` defaults to `<cwd>/.codex`, so a planted rollout sits in the agent's working directory — an eval-awareness hazard specific to synthesizing a prior. Codex warns when the rollout lands inside cwd; pass `home_dir` / `config_dir` to move it out.

## Tests

`tests/test_acp_session_resume.py`, `test_codex_resume.py`, `test_codex_rollout.py`, `test_codex_rollout_messages.py`, `test_claude_code_transcript.py`, `test_claude_code_resume.py` — 86 unit tests over the ACP load path, both agents' resume wiring, and build/parse/ChatMessage round-trips (including real-session edge cases: unmodelled rows, withheld-plaintext reasoning, multimodal blocks).

`tests/test_session_resume_e2e.py` — two tiers against a real sandbox:
- **load tier** (docker, no model call): the synthesized session file is at the exact path the CLI reads, carries the prior, and the spawned CLI shows `--resume=<planted id>` — plus `--resume-session-at=<uuid>` for the truncation case. 4 passing.
- **turn tier** (`--runapi`): one prompt, then assert the prior appears in what the CLI sends upstream and that the model answers from it.

Locally green: `ruff check`, `ruff format --check`, `mypy` (strict, src + tests), `pytest`.

## Dependency

The Claude Code tests need the fix for #96 to run on a macOS/Windows host: without `--libc`, the adapter bundle ships without its native CLI and every session — new or loaded — dies with a bare "Internal error". That fix is a separate branch (`--libc` + cache-key bump, the JSON-RPC `data` detail #96 mentions, and an install-time guard).

## Not included

Gemini CLI gets resume-by-id via the generic path, but not a synthetic prior — that needs its own on-disk session format.


BorisTheBrave:
I'd be interested in seeing this feature for claude code. Currently I'm doing something similar, but via the CLI.

jjallaire:
One question I have here is why not have the rollout specified as a list of `ChatMessage` (versus having to construct the code_cli specific artifact)

BorisTheBrave:
> why not have the rollout specified as a list of ChatMessage 

For me: the session might be sourced from outside inspect. There's also fidelity of representation concerns.

jannes-apollo:
>  One question I have here is why not have the rollout specified as a list of ChatMessage (versus having to construct the code_cli specific artifact)

Changed the code so that ChatMessage is now the main interface.

I kept the typed items underneath because ChatMessage can't express things such as custom_tool_call's free-form input (apply_patch), reasoning plaintext (your Responses converter zeroes it, which is right for an API request but wrong for on-disk history), and rows with no Inspect equivalent (web_search_call, tool_search_call).

The obvious move was to just call openai_responses_inputs, but that imports the openai SDK, which isn't a runtime dep of inspect_swe, so I mirrored that mapping here instead. Happy to swap to it, or to a provider-free version exported from inspect_ai, if you'd prefer either

dragonstyle (dismissed):
Thanks so much for this PR. It looks like there are still some failures in Python 3.10. Could you have a look at these?

jannes-apollo:
I fixed the python 3.10 failures

src/inspect_swe/acp/_agents/claude_code/transcript.py
```diff
@@ -0,0 +1,576 @@
...
+    if isinstance(message.content, str):
+        return [UserText(text=message.content)]
+    items: list[TranscriptItem] = []
+    for block in message.content:
+        if isinstance(block, ContentText):
+            items.append(UserText(text=block.text))
+        else:
+            # Images and other modalities are valid Anthropic blocks; pass the
+            # model_dump through rather than inventing a typed shape for each.
+            items.append(RawBlock(role="user", block=block.model_dump()))
```
craig-meridian, line 448: **[P1]** This `RawBlock(role=..., block=block.model_dump())` fallback is wrong for every non-text Inspect `Content`, not just images. `message.content` here is consumed as raw Anthropic content (per the module docstring), but `model_dump()` produces Inspect's internal shape — e.g. `ContentImage` -> `{"type":"image","image":...,"detail":...}` with no `source`, which I confirmed fails Anthropic `ImageBlockParam` validation (`source` is Required), so Claude rejects the replayed message on the next turn. `ContentDocument` dumps are equally malformed. Same fallback at L471 (`_assistant_items`). The unit test only asserts the value becomes a `RawBlock`, so it stays green over the broken artifact. Note `RawBlock` is doing double duty: it is correct in the `parse_transcript`->`_block_to_item` direction (where `block` is a genuine Anthropic block) but wrong here. **Minimal fix:** raise on non-text content exactly as codex's `_user_items` does (and fix the test); the full data-URI -> `source` conversion can be a follow-up.

src/inspect_swe/acp/_agents/claude_code/transcript.py
```diff
@@ -0,0 +1,576 @@
...
+    reasoning came from an Anthropic endpoint with its signature intact
+    (``ContentReasoning.signature``).
+
+    Also lossy in that system messages have no transcript row — Claude Code
+    takes its system prompt from the CLI, not the session file — so they are
+    skipped; pass them via ``system_prompt`` on the agent instead.
+    """
+    items: list[TranscriptItem] = []
+    for message in messages:
+        if isinstance(message, ChatMessageSystem):
```
craig-meridian, line 338: **[P1]** `resume_messages` is advertised as the generic `ChatMessage` front door, but this silently discards every `ChatMessageSystem`. `_start_agent` derives `CLAUDE_CODE_APPEND_SYSTEM_PROMPT` only from the current `AgentState` and explicit `system_prompt`, not from `resume_messages` — so a saved Inspect trajectory containing task or safety instructions resumes under different instructions (natural when `resume_messages = prior_state.messages`), and a system-only prior produces an empty transcript file. Codex preserves these as `DeveloperText`, so the drop is also an asymmetry between the two agents. **Minimal fix:** raise (or warn loudly) on system messages here rather than dropping them silently; merging them into the Claude system prompt at startup would be the fuller solution and is fine as a follow-up. (See the non-blocking note in `claude_code.py` on *re-appending* the prompt to an already-persisted session — a fuller fix should distinguish the two cases.)

src/inspect_swe/acp/_agents/claude_code/claude_code.py
```diff
@@ -128,6 +164,16 @@ async def _start_agent(
...
+            # resolve $HOME through a shell exec, the adapter resolves it in its
+            # own process env, and if those disagree the transcript lands where
+            # session/load never looks (which reads as a silently fresh session).
+            if self._config_dir is not None or self.is_resuming:
+                self._resolved_config_dir = await self._resolve_config_dir(sbox)
+                agent_env["CLAUDE_CONFIG_DIR"] = str(self._resolved_config_dir)
+
             # System prompt via env (the ACP adapter will forward to CC)
             resolved_prompt = self._resolve_system_prompt(state)
             if resolved_prompt:
```
craig-meridian, line 179: **[Note — non-blocking]** This resolves and exports the system prompt unconditionally, including for `resume_session_id` / `resume_transcript`. The non-ACP Claude runner deliberately suppresses `--append-system-prompt` on resume (see `_claude_code/claude_code.py` ~L312): the bridge round-trips Claude Code's own full system prompt back into `state.messages` as a `ChatMessageSystem`, so resuming with an `AgentState` carried over from a prior run re-appends the entire prompt on top of CC's base prompt. I'm not asking for a change here because the right behavior genuinely needs thought — blanket suppression on resume would be wrong too (CC doesn't persist the system prompt in the transcript, so resume-by-messages *needs* the append). Flagging it so the interaction is a conscious decision; happy for this to be a follow-up.

src/inspect_swe/acp/_agents/claude_code/transcript.py
```diff
@@ -0,0 +1,576 @@
...
+            continue
+        if isinstance(message, ChatMessageUser):
+            items.extend(_user_items(message))
+        elif isinstance(message, ChatMessageAssistant):
+            items.extend(_assistant_items(message, include_thinking))
+        elif isinstance(message, ChatMessageTool):
+            items.append(
+                ToolResult(
+                    tool_use_id=message.tool_call_id or "",
+                    content=message.error.message if message.error else message.text,
```
craig-meridian, line 348: **[Note — non-blocking]** Non-text tool results are silently emptied: `ChatMessageTool.text` is `""` for image-only content, so a prior containing screenshot-style tool output resumes with empty tool results — silent context loss. The codex converter has the same gap (`rollout.py` L400 — and codex's raise-on-unsupported only covers user/assistant blocks, not tool outputs). This is the same multimodal-conversion territory as the content-block issue above, so I'd treat it as part of that follow-up rather than a blocker; `ToolResult.content` already accepts Anthropic block lists when you get there.

src/inspect_swe/acp/_agents/claude_code/claude_code.py
```diff
@@ -45,7 +56,32 @@ def __init__(
...
+            )
+        self._resume_transcript = resume_transcript
+        self._resume_message_uuid = resume_message_uuid
+        if resume_transcript is not None:
+            kwargs["resume_session_id"] = resume_transcript.session_id
         super().__init__(**kwargs)
+        if resume_message_uuid is not None and not self.is_resuming:
+            raise ValueError(
+                "`resume_message_uuid` truncates a conversation being resumed, so it "
+                "needs one of `resume_session_id`, `resume_messages`, or "
```
craig-meridian, line 82: **[P2]** This error message endorses `resume_messages` as a valid companion to `resume_message_uuid`, but that combination can never work: with `resume_messages` the transcript is built inside `_resolve_resume_session` with fresh `uuid4()` row uuids that are never surfaced to the caller, so any uuid the caller passes cannot match a row and `resumeSessionAt` targets a nonexistent uuid. Please tighten the guard (and message) to `resume_session_id` / `resume_transcript` only — the docstring's "uuids come from `TranscriptSpec.item_uuids` or `ParsedTranscript.item_uuids`" already implies this.

src/inspect_swe/acp/_agents/claude_code/claude_code.py
```diff
@@ -128,6 +164,16 @@ async def _start_agent(
...
 
+            # Resolve CLAUDE_CONFIG_DIR — where Claude Code keeps its sessions,
+            # and so where a resumed transcript has to be written. When resuming
+            # we pin it explicitly even if the caller didn't override it: we
+            # resolve $HOME through a shell exec, the adapter resolves it in its
+            # own process env, and if those disagree the transcript lands where
+            # session/load never looks (which reads as a silently fresh session).
+            if self._config_dir is not None or self.is_resuming:
+                self._resolved_config_dir = await self._resolve_config_dir(sbox)
+                agent_env["CLAUDE_CONFIG_DIR"] = str(self._resolved_config_dir)
```
craig-meridian, line 175: **[P2]** A caller-supplied `env={"CLAUDE_CONFIG_DIR": ...}` is silently trampled here when resuming: the merge at L165 lets it through, but this pin overwrites it, and `_resolve_config_dir` consults only `self._config_dir` / `$HOME/.claude` — never `self.env`. Failing scenario: run 1 (no resume) persists sessions under the env-supplied dir; run 2 adds `resume_session_id` for that session, the pin points the adapter at `$HOME/.claude`, and the session isn't found — exactly the silent-fresh-session failure this pinning comment says it exists to prevent. Small fix: have `_resolve_config_dir` prefer `self.env.get("CLAUDE_CONFIG_DIR")` over the default (explicit `config_dir` still winning).

src/inspect_swe/acp/_agents/codex_cli/codex_cli.py
```diff
@@ -108,6 +128,8 @@ async def _start_agent(
                     cwd=self.cwd,
                 )
             await sandbox_exec(sbox, cmd=f"mkdir -p {codex_home}", user=self.user)
+            # Stash for _prepare_resume (base class calls it after this yields).
+            self._codex_home = codex_home
```
craig-meridian, line 132: **[P2]** Mirror image of the claude config-dir issue: the `agent_env = {"CODEX_HOME": codex_home, ...} | self.env` merge below (L177) lets a caller-supplied `env["CODEX_HOME"]` override the resolved home for the adapter process, while the new `_resolve_resume_session` writes the rollout under the *resolved* `self._codex_home` stashed here. `CodexCli(env={"CODEX_HOME": "/custom"}, resume_messages=...)` writes to `<cwd>/.codex/sessions/...` but codex reads `/custom` → `session/load` can't find it → silent fresh session. The merge order pre-exists this PR, but the resume write-path/read-path split is new. Small fix: honor `self.env["CODEX_HOME"]` when resolving (or reject the override when resuming).

src/inspect_swe/acp/agent.py
```diff
@@ -203,21 +309,19 @@ async def __call__(
...
                         )
 
-                        session = await conn.new_session(
-                            cwd=self.cwd,
-                            mcp_servers=acp_mcp_servers or None,  # type: ignore[arg-type]
-                        )
-
                         self.conn = conn
-                        self.session_id = session.session_id
+                        self.session_id = await self._open_session(
```
craig-meridian, line 322: **[P2]** `self.conn = conn` is assigned before this call, but the clearing `finally` only begins after `_open_session` succeeds. `_open_session` adds several new raise paths (the `loadSession` capability gate, synthetic-session materialization, `load_session` itself); if any of them fires, the agent retains a reference to a connection the context manager then closes — and on a reused agent, `session_id` keeps its stale value too. On main this couldn't happen (`self.conn` was assigned only after `new_session` succeeded). Small fix: move the assignment and the `_open_session` call inside the cleanup `try`/`finally`.

src/inspect_swe/acp/_agents/codex_cli/rollout.py
```diff
@@ -0,0 +1,777 @@
...
+                if b.get("type") == "summary_text"
+            )
+            or None
+        )
+        return Reasoning(
+            text=text,
+            summary=summary,
+            encrypted_content=payload.get("encrypted_content"),
+        )
+    if ptype == "message":
```
craig-meridian, line 669: **[P2 — the docstring is the ask; the fidelity fix is optional]** Collapsing an all-text assistant payload to `AssistantText(text=...)` (L680) keeps only the text, and `_item_payload` re-emits just `{type, role, content}` — dropping `phase`, `id` and `status`. `phase` is real and semantically meaningful: codex-rs's `ResponseItem::Message` carries `phase: Option<MessagePhase>` (`Commentary`/`FinalAnswer`), and codex treats a missing phase as "phase unknown" with legacy-compatibility behavior — so parse -> truncate -> rebuild silently regresses rebuilt rows. The stripping is systemic (`function_call` and `reasoning` likewise drop `id`/`status`), which contradicts this module's "Round-trips losslessly for the modelled item types" docstring. The load-bearing `call_id` *is* preserved, so no hard resume break. **Ask:** correct the docstring to state what is actually preserved. **Consider (non-blocking):** bail to `RawResponseItem` whenever a modelled payload carries keys you can't reproduce — the degrade-to-raw philosophy already used here — which would cover `phase` and future drift in one move.

src/inspect_swe/acp/_agents/codex_cli/rollout.py
```diff
@@ -0,0 +1,777 @@
...
+def _make_turn_context(
+    *, cwd: str, model: str, current_date: datetime
+) -> dict[str, Any]:
+    return {
+        "turn_id": str(uuid.uuid4()),
+        "cwd": cwd,
+        "current_date": current_date.strftime("%Y-%m-%d"),
+        "timezone": "UTC",
+        "approval_policy": "never",
+        "sandbox_policy": {
```
craig-meridian, line 696: **[Q]** This hard-codes `workspace-write` with `network_access: false` (and `reasoning_effort: medium` regardless of `config_overrides`), while the ACP codex agent runs with `danger-full-access`. Does codex derive any resumed-turn policy from the recorded `turn_context` rather than live config? If so, a resumed session would run under a tighter sandbox than a fresh one. If it's purely informational, fine as is — just want to confirm.

src/inspect_swe/acp/_agents/claude_code/claude_code.py
```diff
@@ -158,6 +204,81 @@ async def _start_agent(
...
+        # under the resolved spelling rather than spec.relative_path (built from
+        # the logical cwd on the host, which can't see the sandbox's fs).
+        resolved_cwd = await sandbox_exec(
+            sbox, f"realpath {self.cwd}", user=self.user, cwd=self.cwd
+        )
+        transcript_path = join_path(
+            self._resolved_config_dir,
+            f"projects/{project_slug(resolved_cwd)}/{spec.session_id}.jsonl",
+        )
+        await sbox.write_file(transcript_path, spec.content)
```
craig-meridian, line 261: **[P2]** Codex warns when the synthetic session lands inside the agent's cwd (`_warn_if_visible_to_agent`), but Claude Code has no equivalent — yet `config_dir` defaults to `$HOME/.claude`, so with the default sandbox user (cwd `$HOME`, e.g. the `/root` used in the e2e) the planted transcript sits inside the agent's own working directory and is readable by the agent. Same eval-awareness hazard the PR description flags for codex, and the same handful of lines to add: compare the resolved `projects/<slug>` path against the resolved cwd and warn.

src/inspect_swe/acp/_agents/codex_cli/rollout.py
```diff
@@ -0,0 +1,777 @@
...
+    ContentText,
+)
+from inspect_ai.tool import ToolCall
+from pydantic import BaseModel, Field, ValidationError
+
+# Key under which Inspect's OpenAI Responses provider stashes the reasoning
+# ciphertext for non-redacted reasoning (mirrors the private
+# `inspect_ai.model._openai_responses.REASONING_ENCRYPTED_CONTENT`; importing it
+# would pull the `openai` SDK into a runtime path where it isn't a dependency).
+_REASONING_ENCRYPTED_CONTENT = "reasoning_encrypted_content"
```
craig-meridian, line 54: **[P2]** This mirrors the private `inspect_ai.model._openai_responses.REASONING_ENCRYPTED_CONTENT` (I confirmed it's byte-identical). If that key is renamed upstream, reasoning-ciphertext round-tripping breaks silently (no error, just a failed next turn) — and `tests/test_codex_rollout_messages.py` hard-codes the same string as a second independent copy rather than importing the upstream constant. Small fix: a guard test that imports the real constant and asserts equality — and have it *fail* (not skip) if the import breaks: `inspect_ai` is a runtime dep, so an absent constant is exactly the drift the test exists to catch. Moot if the conversion is upstreamed to a provider-free helper exported from `inspect_ai`, which would be a nice follow-up.

src/inspect_swe/acp/_agents/codex_cli/codex_cli.py
```diff
@@ -108,6 +128,8 @@ async def _start_agent(
                     cwd=self.cwd,
                 )
             await sandbox_exec(sbox, cmd=f"mkdir -p {codex_home}", user=self.user)
+            # Stash for _prepare_resume (base class calls it after this yields).
```
craig-meridian, line 131: Nit: stale name — the base class method is `_resolve_resume_session`, not `_prepare_resume`.

src/inspect_swe/acp/_agents/claude_code/claude_code.py
```diff
@@ -158,6 +204,81 @@ async def _start_agent(
...
+                f"Resume transcript was built for cwd {spec.cwd!r} but this agent runs "
+                f"in {self.cwd!r}. Claude Code locates a session by its cwd, so the "
+                f"two have to match — rebuild the transcript with cwd={self.cwd!r}."
+            )
+        sbox = sandbox_env(self.sandbox)
+        # The SDK resolves symlinks before slugging the cwd, so place the file
+        # under the resolved spelling rather than spec.relative_path (built from
+        # the logical cwd on the host, which can't see the sandbox's fs).
+        resolved_cwd = await sandbox_exec(
+            sbox, f"realpath {self.cwd}", user=self.user, cwd=self.cwd
```
craig-meridian, line 255: Nit: unquoted interpolation into `bash -c` (also `eval echo "{...}"` at L211, `mkdir -p` at L213, and the codex equivalents at `codex_cli.py` L126/L130/L210). A cwd with a space usually fails loudly via `sandbox_exec`, but in the unlucky case where both fragments exist as paths, `realpath` returns two lines and the slug silently goes wrong — a silent fresh session. `shlex.quote` is cheap insurance.

craig-meridian (changes requested):
Thanks for this work and apologies we haven't been more proactive on this. The review is primarily from Claude, with oversight from myself. Please do let me know if you think anything is wrong/unreasonable.

---

Strong PR, and the earlier design question (make `ChatMessage` the front door) is resolved well, with the typed items underneath justified by concrete fidelity gaps.

Keen to see this land, so I've kept the asks small — details are all in the inline comments. In short: two P1s in the `ChatMessage` conversion (both have ~5-line minimal fixes, with fuller solutions fine as follow-ups), a handful of few-line P2s, one docstring correction, and one question. Anything marked **[Note — non-blocking]** or "consider" is take-it-or-leave-it.

Separate from the code: worth confirming merge order vs. the #96 `--libc` fix, since the Claude e2e tier can't go green on a macOS/Windows host without it.

src/inspect_swe/acp/_agents/claude_code/claude_code.py
jannes-apollo: Fixed in 0166cc6: Claude now warns when the planted transcript resolves beneath the agent cwd, with config_dir guidance for moving it outside.

src/inspect_swe/acp/_agents/claude_code/transcript.py
jannes-apollo: Fixed in 0166cc6: ChatMessageSystem is now rejected explicitly instead of being silently dropped from synthesized Claude transcripts.

src/inspect_swe/acp/_agents/claude_code/claude_code.py
jannes-apollo: Fixed in 0166cc6: resume_message_uuid is now accepted only with resume_session_id or resume_transcript, and the public docs and tests match that constraint.

src/inspect_swe/acp/_agents/codex_cli/codex_cli.py
jannes-apollo: Fixed in 0166cc6: Codex home resolution now honors env[CODEX_HOME] and pins that same resolved path for both rollout materialization and the adapter process.

src/inspect_swe/acp/_agents/claude_code/transcript.py
jannes-apollo: Fixed in 0166cc6: non-text tool results now raise for both Claude and Codex instead of silently becoming empty output.

src/inspect_swe/acp/_agents/codex_cli/rollout.py
jannes-apollo: Fixed in 0166cc6: the module documentation now states the actual fidelity boundary and no longer claims lossless round-tripping for modeled items.

src/inspect_swe/acp/_agents/claude_code/claude_code.py
jannes-apollo: Fixed in 0166cc6: shell path interpolation is quoted, config-home expansion no longer uses eval, and expansion is limited to a leading ~ or $HOME.

src/inspect_swe/acp/_agents/claude_code/claude_code.py
jannes-apollo: Fixed in 0166cc6: config_dir wins explicitly, otherwise resolution honors env[CLAUDE_CONFIG_DIR] before falling back to $HOME/.claude, and the resolved value is pinned for the adapter.

src/inspect_swe/acp/_agents/codex_cli/rollout.py
jannes-apollo: Fixed in 0166cc6: a guard test imports Inspect's private REASONING_ENCRYPTED_CONTENT constant and fails if it disappears or differs from the mirrored key.

src/inspect_swe/acp/_agents/claude_code/transcript.py
jannes-apollo: Fixed in 0166cc6: Claude transcript conversion now raises on non-text user and assistant content, and the tests assert rejection instead of accepting malformed RawBlocks.

src/inspect_swe/acp/agent.py
jannes-apollo: Fixed in 0166cc6: connection assignment and session opening are inside the cleanup try/finally, so load failures clear both conn and session_id.

src/inspect_swe/acp/_agents/codex_cli/rollout.py
jannes-apollo: Confirmed against current Codex source for 0166cc6: resumed runtime sandbox and approval policy are built from the live Config. Stored turn_context carries history/reference metadata and prior model information; it does not enforce the resumed turn policy.

src/inspect_swe/acp/_agents/claude_code/claude_code.py
jannes-apollo: Acknowledged in 0166cc6. I left the append behavior unchanged because synthesized resume_messages still need the prompt, while suppressing it for persisted sessions requires distinguishing the two resume modes. This remains follow-up design work as suggested.

src/inspect_swe/acp/_agents/codex_cli/codex_cli.py
jannes-apollo: Fixed in 0166cc6: the stale comment now names _resolve_resume_session.

jannes-apollo:
Sorry, Codex posted the comments without permission

src/inspect_swe/_util/node.py
```diff
@@ -163,6 +197,7 @@ def create_npm_bundle(
             "Please install Node.js (https://nodejs.org/) and ensure npm is on "
             "your PATH."
         )
+    _require_npm_libc_support()
```
craig-meridian, line 200: **[P1]** `_require_npm_libc_support()` currently runs for every cold npm bundle, but only the Claude adapter has libc-gated dependencies (as described in [#96](https://github.com/meridianlabs-ai/inspect_swe/issues/96)). This unnecessarily breaks Codex and Gemini installs on npm <11.5.1, where ignoring `--libc` is harmless. Please make the version guard a per-package option enabled for the Claude adapter on every host OS; Linux-to-musl cross-builds need the protection too.

src/inspect_swe/acp/agent.py
```diff
@@ -115,6 +133,20 @@ def __init__(self, **kwargs: Unpack[ACPAgentParams]) -> None:
...
         self.sandbox = kwargs.get("sandbox")
+        self.resume_session_id = kwargs.get("resume_session_id")
+        self.resume_messages = kwargs.get("resume_messages")
+        if self.resume_session_id is not None and self.resume_messages is not None:
+            raise ValueError(
+                "Pass either `resume_session_id` (re-attach to a session already "
+                "on disk) or `resume_messages` (synthesize a prior conversation), "
+                "not both."
+            )
+        if self.resume_messages is not None and not self.resume_messages:
```
craig-meridian, line 144: **[P2]** This guard runs before agent-specific conversion, but both converters can reduce a non-empty `resume_messages` to no history: Claude drops reasoning-only or empty assistant content, while Codex drops empty assistant content and creates a header-only rollout. Both then effectively resume fresh despite this guard contract. Please consider rejecting zero-item conversion results in both agents’ `_resolve_resume_session()` paths.

craig-meridian (commented):
Thanks for the many improvements here — the earlier findings have been addressed well. There are just a couple of findings left to address (the P1 is the only blocker IMO). The empty-history conversion edge case is something we could have caught in the earlier review, so apologies for only surfacing it now.

PR #71, opened 2026-06-29
Handle claude_code refusals gracefully

brandly:
When the `claude_code` solver runs into a refusal, we'd `raise RuntimeError` which would halt the eval. "Native" solvers in Inspect instead treat refusal as failure and continue on to scoring the sample and to run other samples. With these changes, the `claude_code` solver will now mimic the behavior of other solvers.

### DIFF
--- a/src/inspect_swe/_claude_code/claude_code.py
+++ b/src/inspect_swe/_claude_code/claude_code.py
@@ -13,7 +13,7 @@ from inspect_ai.agent import (
     agent_with,
     sandbox_agent_bridge,
 )
-from inspect_ai.model import ChatMessageSystem, GenerateFilter, Model
+from inspect_ai.model import ChatMessageSystem, GenerateFilter, Model, StopReason
 from inspect_ai.scorer import score
 from inspect_ai.tool import MCPServerConfig, Skill, install_skills, read_skills
 from inspect_ai.util import (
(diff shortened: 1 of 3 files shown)


src/inspect_swe/_claude_code/claude_code.py
```diff
@@ -373,10 +373,19 @@ async def execute(state: AgentState) -> AgentState:
                         cc_debug = store_as(ClaudeCodeDebug) if debug else None
                         stderr_data = ""
                         exit_code = 0
+                        terminal_stop_reason: str | None = None
+                        saw_result_event = False
 
                         async for cc_event in claude_code_event_stream(proc):
                             if isinstance(cc_event, JsonlEvent):
                                 consumer.process_jsonl_line(cc_event.raw)
+                                terminal_stop_reason, saw_result_event = (
```
dragonstyle: I think rather than parsing the json here, this makes more sense in LiveConsumer (new property like `last_stop_reason`). In LiveConsumer, we have a sync for Model Events, so you could just collect the stop reason there without needing the manual parsing or extra flags (be sure to clear it in `reset()`).

src/inspect_swe/_claude_code/claude_code.py
```diff
@@ -396,24 +405,32 @@ async def execute(state: AgentState) -> AgentState:
...
-                                and len(stderr_data.strip()) == 0
-                                and retry_uncaught_errors is not None
-                                and uncaught_error_count < retry_uncaught_errors
+                            # Claude Code exits 1 after Anthropic refusal
+                            # responses even though the refusal has already
+                            # been recorded in bridge.state and can be scored.
+                            if not _is_claude_code_refusal_exit(
+                                exit_code,
+                                stderr_data,
+                                terminal_stop_reason,
```
dragonstyle: this can then pass `consumer.last_stop_reason` and check that.

dragonstyle (changes requested):

brandly:
@dragonstyle that cleaned things up nicely!!

dragonstyle (approved):
👍

PR #73, opened 2026-07-01
feat(kimi): add `kimi_code()` agent with Inspect bridging, MCP, skills, and Centaur support

### LINKED ISSUE #72: Support for Kimi Code
https://github.com/MoonshotAI/kimi-code

jannes-apollo:
## Summary

Adds a `kimi_code()` agent for Moonshot AI's [Kimi Code](https://github.com/MoonshotAI/kimi-code) CLI, running headless in a sandbox with Inspect model bridging — the same shape as the existing `opencode()` / `claude_code()` agents. Closes #72.

Kimi's model calls are routed through a generated `~/.kimi-code/config.toml` whose `openai`-protocol provider points at the Inspect bridge, so the returned transcript is the bridge's record of those calls (Kimi CLI stdout is never parsed). MCP servers are wired natively via `~/.kimi-code/mcp.json`.

## What's included

**`_kimi_code/kimi_code.py`** — the agent. Writes `config.toml` (bridge provider + permission rules) and `mcp.json`, prepends system messages to the prompt (kimi `-p` has no separate system-prompt flag), runs `kimi --output-format text [--continue] -p <prompt>`, and returns `bridge.state`. Full parity with the other agents:
- `system_prompt`, `mcp_servers`, `bridged_tools`, `filter`, `model` / `model_aliases`, `retry_refusals`, `cwd` (via the shared `resolve_agent_cwd`), `env`, `user`, `sandbox`.
- `skills` — installed into `~/.kimi-code/skills` and surfaced through kimi's `extra_skill_dirs` config.
- `centaur` — exposes a `kimi` alias to an Inspect `human_cli()` agent.
- `attempts` — multiple scored attempts with `--continue` between them.
- `disallowed_tools` — denied via kimi's native `[[permission.rules]]` (deny wins).

**`_kimi_code/agentbinary.py`** — binary download via the shared `AgentBinarySource`, so `version` supports `auto` / `sandbox` / `latest` / `x.x.x`. Kimi publishes a `latest.json` version pointer and a per-version `binaries/<version>/manifest.json` carrying each platform's filename + sha256, which drives resolution and checksum verification.

**Two Kimi-specific fixups** (in `kimi_code.py`):
- `_dedupe_tool_call_ids` — kimi assigns tool_call_ids like `functions_Bash_0` whose counter resets, so ids recur across a session. The bridge records them verbatim, so a viewer that threads on tool_call_id collapses every reuse into one node (a linear chain renders as a spurious tree — same class as #33). Rewrites ids to be globally unique post-generation while preserving assistant↔tool pairing.
- Repeat-reminder strip — kimi's CLI injects a `<system-reminder>Repeated tool call detected…</system-reminder>` nag. Re-polling the same read tool is legitimate in long-horizon tasks, so this strips the nag before the bridged model sees it. Composed with any user-supplied `filter`. **This one is mildly opinionated** — happy to gate it behind an option or drop it if you'd prefer.

Also registered in `__init__.py` / `_registry.py`, `docs/kimi_code.qmd` (+ nav + index), and `CHANGELOG.md`.

## Tests

`tests/test_kimi_setup.py` — 10 tests: version resolution (latest pointer → manifest, and specific-version), platform mapping (incl. musl→glibc collapse) and unknown-platform error, `config.toml` / `mcp.json` generation, and both fixups (reminder strip preserving pairing, tool_call_id dedupe preserving FIFO pairing). Includes a network-gated live drift-check asserting kimi's `latest.json` + `manifest.json` still expose linux binaries with sha256 checksums.

Locally green: `ruff check`, `ruff format --check`, `mypy src tests`, `pytest tests/test_kimi_setup.py`.

**End-to-end smoke test** (docker sandbox, `python:3.12-bookworm`, `version="latest"`, bridged to `anthropic/claude-sonnet-4-5`): a one-sample task asking the agent to write `/root/answer.txt`. The agent downloaded the current binary (0.21.1), ran headless, used its native `Write` tool through the bridge, and the file-based scorer confirmed the contents — `accuracy 1.000`. Transcript threading and the tool_call_id dedupe both applied correctly.

## Notes / limitations

- The runtime path is adapted from an implementation that's been running in production RL rollouts. The full agent — including the new **upstream binary-download path** (`agentbinary.py`) — was exercised end-to-end via the smoke test above; it has not been run at scale or against a broad task suite.
- kimi ships glibc linux builds keyed by arch only (no musl variants), so musl sandbox platforms are collapsed onto the plain `linux-x64` / `linux-arm64` keys.

### DIFF
--- a/docs/kimi_code.qmd
+++ b/docs/kimi_code.qmd
@@ -22,6 +22,7 @@ The following options are supported for customizing the behavior of the agent:
 | `centaur` | Run in [Centaur Mode](#centaur-mode), which makes {{< meta agent_name >}} available to an Inspect `human_cli()` agent rather than running it unattended. |
 | `attempts` | Allow the agent to have multiple scored attempts at solving the task. |
 | `model` | Model name to use for agent (defaults to main model for task). |
+| `max_context_size` | Context window to configure for Kimi. Defaults to the bridged model's context length from Inspect model metadata; required when that metadata is unavailable. |
 | `filter` | Filter for intercepting bridged model requests. |
 | `retry_refusals` | Should refusals be retried? (pass number of times to retry) |
 | `disallowed_tools` | Tool names to deny via {{< meta agent_name >}} permission rules. |
--- a/docs/kimi_code.qmd
+++ b/docs/kimi_code.qmd
@@ -43,7 +43,13 @@ For example, here we specify a custom system prompt:
 
 {{< include _snippets/agent_bridged_tools.md >}}
 
+## Tool Repetition Reminders
+
+The {{< meta agent_name >}} CLI appends escalating `<system-reminder>` notes to tool results when it detects the same tool call repeated several times in a row. Since re-polling an identical read-only command is often legitimate in long-running tasks (e.g. waiting for a build to finish), the `kimi_code()` agent strips the first two escalation tiers of these reminders before the bridged model sees them (so they also don't appear in the transcript). The final tier — which instructs the model to write its final response, and precedes {{< meta agent_name >}} force-stopping the turn at 12 repetitions — is passed through unchanged so the model can wrap up gracefully.
+
 ## Installation
+
+{{< meta agent_name >}} publishes glibc-only Linux binaries, and the agent runs them via `bash`, so sandbox images must provide glibc and bash (e.g. Debian/Ubuntu based images). musl-based images such as Alpine are not supported and fail at binary resolution time.
 
 By default, the agent will download the current latest version of {{< meta agent_name >}} and copy it to the sandbox. You can override this behaviour using the `version` option:
 
--- a/docs/kimi_code.qmd
+++ b/docs/kimi_code.qmd
@@ -1,0 +1,63 @@
+---
+title: Kimi Code
+agent: kimi_code
+agent_provider: "Moonshot AI"
+agent_name: "Kimi Code"
+agent_url: "https://github.com/MoonshotAI/kimi-code"
+agent_disallowed_tool: "WebFetch"
+---
+
+{{< include _snippets/agent_intro.md >}}
+
+## Options
+
+The following options are supported for customizing the behavior of the agent:
+
+| Option | Description |
+|------------------------------------|------------------------------------|
+| `system_prompt` | Additional system prompt to append to default system prompt. |
+| `skills` | Additional [skills](https://inspect.aisi.org.uk/tools-standard.html#sec-skill) to make available to the agent. |
+| `mcp_servers` | MCP servers (see [MCP Servers](#mcp-servers) below for details). |
+| `bridged_tools` | Host-side Inspect tools to expose via MCP (see [Bridged Tools](#bridged-tools) below for details). |
+| `centaur` | Run in [Centaur Mode](#centaur-mode), which makes {{< meta agent_name >}} available to an Inspect `human_cli()` agent rather than running it unattended. |
+| `attempts` | Allow the agent to have multiple scored attempts at solving the task. |
+| `model` | Model name to use for agent (defaults to main model for task). |
+| `filter` | Filter for intercepting bridged model requests. |
+| `retry_refusals` | Should refusals be retried? (pass number of times to retry) |
+| `disallowed_tools` | Tool names to deny via {{< meta agent_name >}} permission rules. |
+| `cwd` | Working directory for {{< meta agent_name >}} session. |
+| `env` | Environment variables to set for {{< meta agent_name >}}. |
+| `version` | Version of {{< meta agent_name >}} to use (see [Installation](#installation) below for details) |
+
+: {tbl-colwidths=\[25,75\]}
+
+For example, here we specify a custom system prompt:
+
+```python
+{{< meta agent >}}(
+    system_prompt="You are an ace system researcher."
+)
+```
+
+{{< include _snippets/agent_mcp_servers.md >}}
+
+{{< include _snippets/agent_bridged_tools.md >}}
+
+## Installation
+
+By default, the agent will download the current latest version of {{< meta agent_name >}} and copy it to the sandbox. You can override this behaviour using the `version` option:
+
+| Option | Description |
+|------------------------------------|------------------------------------|
+| `"auto"` | Use any available version of {{< meta agent_name >}} in the sandbox, otherwise download the latest version. |
+| `"sandbox"` | Use the version of {{< meta agent_name >}} in the sandbox (raises `RuntimeError` if not available in the sandbox) |
+| `"latest"` | Download and use the very latest version. |
+| `"x.x.x"` | Download and use a specific version number. |
+
+: {tbl-colwidths=\[25,75\]}
+
+{{< include _snippets/agent_installation.md >}}
+
+{{< include _snippets/agent_centaur.md >}}
+
+{{< include _snippets/agent_troubleshooting.md >}}
--- a/src/inspect_swe/_kimi_code/agentbinary.py
+++ b/src/inspect_swe/_kimi_code/agentbinary.py
@@ -1,0 +1,76 @@
+import json
+from pathlib import Path
+from typing import Any
+
+from typing_extensions import Literal
+
+from .._util.agentbinary import AgentBinarySource, AgentBinaryVersion
+from .._util.appdirs import package_cache_dir
+from .._util.download import download_text_file
+from .._util.sandbox import SandboxPlatform
+
+# Kimi Code publishes raw per-platform binaries plus a version pointer and a
+# per-version manifest of filenames and sha256 checksums.
+KIMI_DIST_URL = "https://code.kimi.com/kimi-code/binaries"
+KIMI_LATEST_URL = "https://code.kimi.com/kimi-code/latest.json"
+
+
+def kimi_code_binary_source() -> AgentBinarySource:
+    cached_binary_dir = package_cache_dir("kimi-code-downloads")
+
+    async def resolve_version(
+        version: Literal["stable", "latest"] | str, platform: SandboxPlatform
+    ) -> AgentBinaryVersion:
+        if version in ["stable", "latest"]:
+            version = await _fetch_latest_version()
+
+        manifest = await _fetch_manifest(version)
+        platforms = manifest["platforms"]
+        kimi_platform = _platform_to_kimi_platform(platform)
+        if kimi_platform not in platforms:
+            raise RuntimeError(
+                f"No Kimi Code binary for platform {platform} in version {version}"
+            )
+        entry = platforms[kimi_platform]
+        download_url = f"{KIMI_DIST_URL}/{version}/{entry['filename']}"
+        return AgentBinaryVersion(version, entry["checksum"], download_url)
+
+    def cached_binary_path(version: str, platform: SandboxPlatform) -> Path:
+        return cached_binary_dir / f"kimi-code-{version}-{platform}"
+
+    def list_cached_binaries() -> list[Path]:
+        return list(cached_binary_dir.glob("kimi-code-*"))
+
+    return AgentBinarySource(
+        agent="kimi code",
+        binary="kimi",
+        resolve_version=resolve_version,
+        cached_binary_path=cached_binary_path,
+        list_cached_binaries=list_cached_binaries,
+        post_download=None,
+        post_install=None,
+    )
+
+
+def _platform_to_kimi_platform(platform: SandboxPlatform) -> str:
+    # Kimi ships glibc linux builds keyed by arch only (no musl variants), so
+    # collapse musl platforms onto the plain linux key.
+    platform_map = {
+        "linux-x64": "linux-x64",
+        "linux-x64-musl": "linux-x64",
+        "linux-arm64": "linux-arm64",
+        "linux-arm64-musl": "linux-arm64",
+    }
+    if platform not in platform_map:
+        raise ValueError(f"Unsupported platform: {platform}")
+    return platform_map[platform]
+
+
+async def _fetch_latest_version() -> str:
+    latest = json.loads(await download_text_file(KIMI_LATEST_URL))
+    return str(latest["version"])
+
+
+async def _fetch_manifest(version: str) -> dict[str, Any]:
+    manifest_url = f"{KIMI_DIST_URL}/{version}/manifest.json"
+    return dict(json.loads(await download_text_file(manifest_url)))
--- a/src/inspect_swe/_kimi_code/agentbinary.py
+++ b/src/inspect_swe/_kimi_code/agentbinary.py
@@ -53,17 +53,16 @@ def kimi_code_binary_source() -> AgentBinarySource:
 
 
 def _platform_to_kimi_platform(platform: SandboxPlatform) -> str:
-    # Kimi ships glibc linux builds keyed by arch only (no musl variants), so
-    # collapse musl platforms onto the plain linux key.
-    platform_map = {
-        "linux-x64": "linux-x64",
-        "linux-x64-musl": "linux-x64",
-        "linux-arm64": "linux-arm64",
-        "linux-arm64-musl": "linux-arm64",
-    }
-    if platform not in platform_map:
-        raise ValueError(f"Unsupported platform: {platform}")
-    return platform_map[platform]
+    # Kimi publishes glibc-only linux builds (no musl variants), so there is no
+    # binary to fall back to on musl images (e.g. Alpine) — the glibc build
+    # would download fine and then fail at exec with a cryptic "required file
+    # not found". Raise at resolution time instead.
+    if platform.endswith("-musl"):
+        raise RuntimeError(
+            f"Kimi Code publishes glibc-only linux builds; no musl binary is "
+            f"available for {platform}."
+        )
+    return platform
 
 
 async def _fetch_latest_version() -> str:
--- a/src/inspect_swe/_kimi_code/kimi_code.py
+++ b/src/inspect_swe/_kimi_code/kimi_code.py
@@ -15,6 +15,7 @@ from inspect_ai.agent import (
     agent_with,
     sandbox_agent_bridge,
 )
+from inspect_ai.agent._bridge.util import resolve_inspect_model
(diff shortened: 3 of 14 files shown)


src/inspect_swe/_kimi_code/kimi_code.py
```diff
@@ -0,0 +1,460 @@
...
+]
+
+KIMI_MAX_CONTEXT_SIZE = 262144
+
+# Kimi Code's CLI injects a <system-reminder> nagging the model not to repeat an
+# identical tool call. Re-polling the same read tool with the same arguments is
+# legitimate (a build hasn't finished, a status is still pending), so the nag is
+# a harness artifact rather than part of the task; strip it before the bridged
+# model sees it.
+_REPEAT_REMINDER_MARKER = "Repeated tool call detected"
```
craig-meridian: >Repeat-reminder strip — kimi's CLI injects a `<system-reminder>Repeated tool call detected…</system-reminder>` nag. Re-polling the same read tool is legitimate in long-horizon tasks, so this strips the nag before the bridged model sees it. Composed with any user-supplied `filter`. **This one is mildly opinionated** — happy to gate it behind an option or drop it if you'd prefer.

I don't have an issue with this, and we can always add an option later on if some user's don't like it. But I think we should document this in the kimi_code.qmd doc page so make sure users are aware of this.

Another option which you've touched upon if to add an option to `kimi_code()` with a suitable docstring, but I'm not advocating for this unless you think users are likely to want to disable this.

uv.lock
```diff
@@ -1206,8 +1215,8 @@ wheels = [
 
 [[package]]
 name = "inspect-ai"
-version = "0.3.235"
-source = { registry = "https://pypi.org/simple" }
+version = "0.3.245.dev7+g26395f389"
```
craig-meridian: I'm hoping you can revert the changes to `uv.lock`. The `inspect_ai` dependency was [recently updated](https://github.com/meridianlabs-ai/inspect_swe/commit/3fc3a848ad9eb76aca8b0d1fdd41f6e5faa51859) to `inspect_ai>=0.3.244` on `main` - hopefully this is sufficient?

In your branch, the `uv.lock` (Inspect via PyPI) conflicts with the `pyproject.toml` (inspect via git). On main, they're both from PyPI.

Lmk if you needed this particular version of Inspect though.

src/inspect_swe/_kimi_code/kimi_code.py
```diff
@@ -0,0 +1,460 @@
...
+                f"kimi_code only supports stdio MCP servers, got {type(server).__name__}"
+            )
+        entry: dict[str, object] = {
+            "command": server.command,
+            "args": list(server.args),
+        }
+        if server.env:
+            entry["env"] = dict(server.env)
+        servers[server.name] = entry
+    return json.dumps({"mcpServers": servers}, indent=2)
```
craig-meridian, line 466: On `MCPServerConfigStdio`, we're not considering whether `cwd` has been set. Consider:
```py
if server.cwd is not None:
    entry["cwd"] = str(server.cwd)
```

We're also ignoring the value of `MCPServerConfig.tools`. Either support it, or raise an error if it is not equal to `"all"`.

docs/kimi_code.qmd
```diff
@@ -0,0 +1,63 @@
...
+
+For example, here we specify a custom system prompt:
+
+```python
+{{< meta agent >}}(
+    system_prompt="You are an ace system researcher."
+)
+```
+
+{{< include _snippets/agent_mcp_servers.md >}}
```
craig-meridian, line 43: I believe that this imported doc will imply that `MCPServerConfigHTTP` is supported (as well as `MCPServerConfigStdio`). We should call out the fact that `MCPServerConfigHTTP` isn't (yet) supported for Kimi Code.

src/inspect_swe/_kimi_code/agentbinary.py
```diff
@@ -0,0 +1,76 @@
...
+    # collapse musl platforms onto the plain linux key.
+    platform_map = {
+        "linux-x64": "linux-x64",
+        "linux-x64-musl": "linux-x64",
+        "linux-arm64": "linux-arm64",
+        "linux-arm64-musl": "linux-arm64",
+    }
+    if platform not in platform_map:
+        raise ValueError(f"Unsupported platform: {platform}")
+    return platform_map[platform]
```
craig-meridian: (heavily assisted by Claude Code, please push back if you disagree)

`_platform_to_kimi_platform` collapses `linux-*-musl` onto the glibc `linux-*` key. But Kimi publishes **glibc-only** linux builds — the real `kimi-code-linux-x64` is `dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2` and needs GLIBC ≤ 2.28, none of which exist on musl. So on an Alpine sandbox the agent downloads ~158 MB, `chmod +x`s it, and only *then* fails at exec with a cryptic error that never mentions libc.
Reproduced end-to-end through the harness (`kimi_code(version="latest")`, same task, only the base image differs):
| Sandbox libc | Exit | stderr | Binary ran? |
|---|---|---|---|
| musl (alpine) | **127** | `kimi-…-linux-arm64-musl: cannot execute: required file not found` | ❌ never starts |
| glibc (bookworm) | — | starts, reaches the bridge (matches your `accuracy 1.000` smoke test) | ✅ |

Since there's no musl build to fall back to, I'd suggest **raising at resolution time** rather than silently remapping — e.g. in `_platform_to_kimi_platform`:

```python
if platform.endswith("-musl"):
    raise RuntimeError(
        f"Kimi Code publishes glibc-only linux builds; no musl binary is "
        f"available for {platform}."
    )
```

craig-meridian: Orthogonal, but `python:3.12-alpine` ships no bash, and `resolve_agent_cwd` runs `bash -c "pwd"`, so bare-Alpine actually dies there first. Not necessarily this PR's job to fix, but it means musl images are doubly unsupported today — a documented "glibc + bash required" note might save someone a confusing debug session.

src/inspect_swe/_kimi_code/kimi_code.py
```diff
@@ -0,0 +1,460 @@
...
+
+        async with sandbox_agent_bridge(
+            state,
+            model=bridge_model,
+            model_aliases=model_aliases,
+            filter=combined_filter,
+            sandbox=sandbox,
+            retry_refusals=retry_refusals,
+            port=port,
+            bridged_tools=bridged_tools,
```
craig-meridian, line 234: Passing `bridged_tools` to `kimi_code()` raises before the agent runs.

The bridge exposes host-side tools to the sandboxed CLI as an **HTTP MCP endpoint it hosts itself**. That's the only transport it offers. So `sandbox_agent_bridge` populates `bridge.mcp_server_configs` with `MCPServerConfigHTTP` instances (see `_register_bridged_tools` in inspect_ai, and the field is typed `list[MCPServerConfigHTTP]`).

We then merge those into `all_mcp_servers` and hand them to `_mcp_json`, which rejects anything non-stdio.

`gemini_cli` and `opencode` handle this by translating the bridge's HTTP configs into their CLI's native remote-MCP form (httpUrl / type: "remote") rather than rejecting them. kimi_code needs the same, or, if Kimi has no HTTP/remote MCP transport, bridged_tools should be rejected at the API boundary with a clear error instead of crashing mid-setup.

src/inspect_swe/_kimi_code/kimi_code.py
```diff
@@ -0,0 +1,460 @@
...
+
+    resolved_disallowed = list(disallowed_tools or [])
+
+    async def execute(state: AgentState) -> AgentState:
+        # determine port (use new port for each execution of agent on sample)
+        MODEL_PORT = "kimi_code_model_port"
+        port = store().get(MODEL_PORT, 3100) + 1
+        store().set(MODEL_PORT, port)
+
+        async def combined_filter(
```
craig-meridian, line 198: This doesn't support the legacy/deprecated str-first filters. `GenerateFilter` ([src](https://github.com/UKGovernmentBEIS/inspect_ai/blob/main/src/inspect_ai/model/_model.py#L1657)) is still a union of Model-first and str-first callables, and the bridge dispatches by inspecting the user filter's first-parameter annotation. Wrapping the user filter inside Model-first `combined_filter` and calling it via `cast(_ModelFilter, filter)(model, …)` means a deprecated-but-supported async def f(model: str, …) filter receives a Model instance so `model.startswith(...)` raises `AttributeError` on every generate call. The same filter works with all sibling agents, which pass filter straight to the bridge.

src/inspect_swe/_kimi_code/kimi_code.py
```diff
@@ -0,0 +1,460 @@
...
+
+            cmd = [kimi_binary, "--output-format", "text"]
+
+            agent_env = {
+                "KIMI_CODE_HOME": kimi_home,
+                "HOME": sandbox_home,
+            } | (env or {})
+
+            if centaur:
+                await _run_kimi_code_centaur(
```
craig-meridian, line 298: Centaur mode can't start: `--output-format text` is baked into the interactive alias

`cmd = [kimi_binary, "--output-format", "text"]` is built before this centaur branch, and `_run_kimi_code_centaur` aliases `kimi` to that full command. But kimi rejects `--output-format` outside prompt mode, so when a human operator types `kimi` they get an immediate error instead of an interactive session.

Verified against the real binary (kimi 0.23.6, linux-arm64, in a container):

```sh
$ kimi --output-format text # the centaur alias form (no -p)
error: Output format is only supported in prompt mode. # exit 1

$ kimi -p "hi" --output-format text # control: flag is fine WITH -p
error: failed to run prompt: No model configured ... # exit 1 (parsing OK)
```

Fix: only append `--output-format text` in the non-centaur branch, the way `claude_code` gates its `--print/--output-format` flags behind `if centaur is False`.

src/inspect_swe/_kimi_code/kimi_code.py
```diff
@@ -0,0 +1,460 @@
...
+    port: int,
+    mcp_servers: Sequence[MCPServerConfig],
+    disallowed_tools: Sequence[str],
+    extra_skill_dirs: Sequence[str],
+    skip_afk_prompt_injection: bool = False,
+) -> str:
+    lines = ['default_model = "bridge"']
+    if extra_skill_dirs:
+        dirs = ", ".join(f'"{d}"' for d in extra_skill_dirs)
+        lines.append(f"extra_skill_dirs = [{dirs}]")
```
craig-meridian, line 509: There are a few places we interpolate user-supplied strings into the TOML file here. We should make use of the existing `from .._util.toml import _format_value` function so that unexpected user-supplied strings don't break the TOML.

src/inspect_swe/_kimi_code/kimi_code.py
```diff
@@ -0,0 +1,460 @@
...
+                            bridge.state, answer_scores
+                        )
+                    else:
+                        agent_prompt = attempts.incorrect_message
+
+                if debug:
+                    debug_output.insert(0, "Kimi Code Debug Output:")
+                    trace("\n".join(debug_output))
+
+        _dedupe_tool_call_ids(bridge.state.messages)
```
craig-meridian: **`_dedupe_tool_call_ids` runs at the wrong layer — the #33 artifact it targets will still be in the log**

`_dedupe_tool_call_ids(bridge.state.messages)` runs *after* the `async with sandbox_agent_bridge(...)` block exits, so it only rewrites the final returned `AgentState`. By then every `ModelEvent` has already been recorded, each with kimi's original colliding ids. The transcript/events view renders from those recorded `ModelEvent`s, so the spurious-tree artifact this code is meant to fix is still present in the eval log — only the returned-state copy ends up clean.

inspect_ai is explicit that a fix which must land in *both* the events and the state has to happen at the `bridge_generate` filter chokepoint, before recording — quoting `agent/_bridge/util.py`:

> mutating the messages in place so the restored source persists in BOTH the recorded `ModelEvent` input and `bridge.state.messages` (and thus the eval log and the ACP TUI)

(the bridge does exactly this for operator-message provenance). The filter runs on `input_messages` immediately before `model.generate(input=input_messages, ...)`, so a rewrite there flows into both.

There's also a live-path angle the post-hoc pass can't cover: because kimi's counter resets, a long enough session can replay two tool calls with the same id *within a single request*. That duplicate is sent to the backing provider during the run, and the Anthropic path maps `tool_use_id` verbatim — so it can be rejected or mispaired mid-session, before any post-hoc cleanup. (The smoke test passing just means no reset happened to collide within one request in that short run.)

Suggested fix: do the FIFO id-rewrite inside `combined_filter` (pre-generation), where `_strip_repeat_reminders` already lives, instead of on `bridge.state.messages` afterward. A per-request pass makes each outgoing request internally unique and lands the deduped ids in the recorded events and the final state in one place.

src/inspect_swe/_kimi_code/kimi_code.py
craig-meridian: It looks like kimi-code will output a different message when it detects a different number of repeated tool calls and this string doesn't appear to catch them all. Either we should catch them all, or none of them. https://github.com/MoonshotAI/kimi-code/blob/main/packages/agent-core-v2/src/agent/toolDedupe/toolDedupeService.ts

tests/test_kimi_setup.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def test_platform_musl_collapses_to_glibc_key() -> None:
+    assert agentbinary._platform_to_kimi_platform("linux-x64-musl") == "linux-x64"
+    assert agentbinary._platform_to_kimi_platform("linux-arm64-musl") == "linux-arm64"
+
+
+def test_resolve_version_unknown_platform_raises() -> None:
+    source = agentbinary.kimi_code_binary_source()
+    with patch.object(
```
craig-meridian, line 92: I _think_ this outer patch might be redundant?

craig-meridian (changes requested):
Thanks for this great contribution Jannes. Looks very promising.

I've reviewed this with the help of Claude Code and left some comments. I've reviewed all of its findings and they appear valid to me, but please do feel free to push back if you disagree.

jannes-apollo:
Thanks for all the feedback, will look into it and update the code

uv.lock
jannes-apollo: I reverted it

src/inspect_swe/_kimi_code/kimi_code.py
jannes-apollo: I added support

tests/test_kimi_setup.py
jannes-apollo: removed it

src/inspect_swe/_kimi_code/kimi_code.py
```diff
@@ -0,0 +1,548 @@
...
+    lines += [
+        "",
+        "[providers.bridge]",
+        'type = "openai"',
+        f'base_url = "http://localhost:{port}/v1"',
+        'api_key = "sk-none"',
+        "",
+        "[models.bridge]",
+        'provider = "bridge"',
+        'model = "inspect"',
```
craig-meridian: **`model=` is silently ignored**

Sorry, I missed this in my initial review. The generated Kimi configuration always requests the literal `"inspect"` model. The `model=bridge_model` value passed to `sandbox_agent_bridge()` does not override requests for the `"inspect"` sentinel (see `resolve_inspect_model` in `inspect_ai/agent/_bridge/util.py` — the fallback is skipped exactly when the requested name is `"inspect"` and the fallback is `inspect/`-prefixed), so `kimi_code(model="openai/gpt-5")` still routes to the task's default model.

This can also produce a mismatch where `max_context_size` is derived from the explicitly requested model while generation is performed by the task model.

Could we write the requested model name into this stanza—effectively `model or "inspect"`, using the bare name rather than the `inspect/`-prefixed form, since alias resolution happens on the raw requested name before prefix stripping—and ensure the context-size lookup resolves the same target, including `model_aliases`? It would be good to add tests covering the default model, an explicit model, and an explicit alias. An unresolvable explicit model should fail clearly rather than silently falling back.

craig-meridian (changes requested):
Thanks a lot for thoroughly addressing my review comments.

Apologies for spotting one more issue which I missed on the first pass.

Once that + fixing failing CI checks are addressed, I think we'll be in a position to merge.

src/inspect_swe/_kimi_code/kimi_code.py
jannes-apollo: Should be fixed now

craig-meridian (approved):
Thanks a lot for this. I've taken the liberty of pushing 3 commits:
- Remove changelog entry (because of merged w/ main it was now under a released section, and we now automate changelog entires with Release Please via PR title)
- Import GenerateInput from the public package (not private underscore-prefixed one).
- A small fix; see below.

**Reminder strip doesn't persist into the returned `AgentState`**
`combined_filter` returns the cleaned messages via `GenerateInput`. In `bridge_generate` (inspect_ai `_bridge/util.py`), a `GenerateInput` result only rebinds a local — the caller in `completions.py` still passes the *original* list to `bridge._track_state()`. So the bridged model (and the recorded `ModelEvent` input) correctly never sees the tier 1–2 reminders, but `bridge.state.messages` — the `AgentState` that `kimi_code()` returns — still contains them, which contradicts the "(so they also don't appear in the transcript)" note in `kimi_code.qmd`. Your own comment on `_dedupe_tool_call_ids` ("in place (not via GenerateInput) so the rewritten ids persist…") describes exactly this mechanism; the strip just didn't take the same route, I'm guessing because it can drop whole messages.

The fix is small — persist the cleanup by updating the list in place (`messages[:] = cleaned`) inside `_strip_repeat_reminders`, mirroring the in-place approach `_dedupe_tool_call_ids` already takes. It's safe across refusal retries (the retry loop resets to the same list object, and the strip is idempotent on already-cleaned content).

Since it's a tiny change and you've already turned around several rounds of feedback, I've pushed it to the branch to save you a round trip — but please treat it as a proposal: push back or rework it if you'd approach it differently.

---

Lmk if that looks good, and I'll merge.

jannes-apollo:
lgtm

Issue #62, opened 2026-06-03
Support for antigravity CLI


alexlloyd-apollo:
Gemini CLI is being replaced by Antigravity CLI:

- https://developers.googleblog.com/an-important-update-transitioning-gemini-cli-to-antigravity-cli/
- https://antigravity.google/docs/gcli-migration

PR #85, opened 2026-07-15
feat: add antigravity_cli agent

sjawhar:
## Summary

Adds an `antigravity_cli` agent that runs Google's Antigravity CLI (`agy`, the successor to Gemini CLI) headless inside an Inspect sandbox, with model calls routed through the sandbox agent bridge — mirroring the existing `gemini_cli` agent.

## How it works

Unlike `gemini_cli` (which honors `GOOGLE_GEMINI_BASE_URL`), `agy` has no base-URL/endpoint override, so this agent:

- Provisions `agy` and a mitmproxy into the sandbox at runtime.
- Sets `HTTPS_PROXY` and uses an in-sandbox mitmdump addon to route `agy`'s Vertex `:streamGenerateContent` calls to the bridge's Gemini-compatible endpoint; all other traffic passes through to Google.

Two request normalizations in the addon let `agy` interoperate with the bridge:

- **Tool-schema types:** Vertex emits the JSON-Schema `type` keyword in SCREAMING_SNAKE_CASE (`STRING`, `OBJECT`, …) in tool declarations; the addon lowercases these to the JSON-Schema strings the bridge expects.
- **functionResponse role:** `agy` places `functionResponse` parts (tool results) under `role="model"` rather than the conventional `role="user"`. The bridge maps tool results to tool messages only under `role="user"`, so without normalization `agy`'s own tool results are dropped from the next request and the agent stalls after a single tool call. The addon relabels (and, for mixed content, splits) those parts to `role="user"` so the tool-call loop proceeds.

## Tests

`tests/test_antigravity_cli.py` covers the addon transforms — including the `functionResponse` role normalization (pure and mixed-content cases) — and the agent wiring/registration.

### DIFF
--- a/src/inspect_swe/__init__.py
+++ b/src/inspect_swe/__init__.py
@@ -1,5 +1,6 @@
 from ._claude_code.claude_code import claude_code
 from ._codex_cli.codex_cli import codex_cli
+from ._antigravity_cli.antigravity_cli import antigravity_cli
 from ._gemini_cli.gemini_cli import gemini_cli
 from ._mini_swe_agent.mini_swe_agent import mini_swe_agent
 from ._opencode.opencode import opencode
@@ -25,6 +26,7 @@ __all__ = [
     "bridge_mcp_to_acp",
     "claude_code",
     "codex_cli",
+    "antigravity_cli",
     "gemini_cli",
     "mini_swe_agent",
     "opencode",
--- a/src/inspect_swe/_antigravity_cli/__init__.py
+++ b/src/inspect_swe/_antigravity_cli/__init__.py
@@ -1,0 +1,3 @@
+from .antigravity_cli import antigravity_cli
+
+__all__ = ["antigravity_cli"]
--- a/src/inspect_swe/_antigravity_cli/_addon.py
+++ b/src/inspect_swe/_antigravity_cli/_addon.py
@@ -1,0 +1,197 @@
+"""mitmproxy addon: route agy's Vertex :streamGenerateContent to the Inspect
+bridge model_proxy (dynamic port). Everything else passes through to Google.
+
+Uses ``requests`` (bundled with the mitmproxy standalone binary) run off the
+event loop via ``run_in_executor``, not ``httpx``: the standalone binary's
+embedded Python environment does not ship httpx (or anyio), only requests.
+"""
+
+import asyncio  # noqa: ANYIO_OK -- anyio is not present in mitmdump's embedded env
+import functools
+import json
+import os
+import re
+from typing import TypeAlias
+
+import requests
+from mitmproxy import http
+
+BRIDGE_PORT = int(os.environ.get("ANTIGRAVITY_BRIDGE_PORT", "13131"))
+_MODEL_RE = re.compile(r"models/([^/:]+)")
+
+# Vertex/Gemini's Schema proto uses SCREAMING_SNAKE_CASE type names for the
+# JSON Schema "type" keyword in tool declarations; the bridge's ToolParams
+# model expects lowercase JSON Schema type strings.
+_SCHEMA_TYPE_NAMES = frozenset(
+    {"STRING", "NUMBER", "INTEGER", "BOOLEAN", "ARRAY", "OBJECT", "NULL"}
+)
+
+JsonValue: TypeAlias = (
+    "dict[str, JsonValue] | list[JsonValue] | str | int | float | bool | None"
+)
+
+
+def _lowercase_schema_type(value: JsonValue) -> JsonValue:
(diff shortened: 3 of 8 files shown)


craig-meridian:
Thanks for this, Sami — careful work on a real problem. With Gemini CLI’s consumer path now unavailable and the request in #62, we do want an Antigravity story, and the addon clearly comes from someone who understands the wire format and sandbox model well.

But we're not comfortable with the `agy`-via-mitmproxy approach.

**The blocker:** `provision_antigravity_home` copies the host’s `~/.gemini/antigravity-cli` directory—including `antigravity-oauth-token`—into the sandbox `$HOME`. The evaluated process then runs with unrestricted shell access under the same sandbox identity via `--dangerously-skip-permissions`. It can read the token and attempt to exfiltrate it; in an ordinarily networked sandbox, it can also clear the proxy variables and make authenticated Google requests that never touch the bridge. Isolated egress can block that direct route, but it does not restore the missing credential boundary.

That breaks the invariant upheld by the other `inspect_swe` agents: only dummy credentials enter the sandbox, while real credentials remain in the host-side process. This is not patchable within this architecture because `agy` is both the evaluated process and the process that must possess the credential. There is no separate harness principal to protect it from the agent’s own shell tools.

[[Google has already enforced bans](https://github.com/google-gemini/gemini-cli/discussions/20632)](https://github.com/google-gemini/gemini-cli/discussions/20632) against third-party tools and proxies that piggyback OAuth to access Antigravity resources or quotas. This design is not exactly the abuse pattern Google described—it redirects inference away from Google rather than consuming Google quota through another client—but it is close enough to create account risk that we should not ask users to accept.

On maintenance, the addon couples us to `agy`’s internal Vertex host and path, request format, and the `functionResponse`/schema-type quirks. The CLI releases frequently and does publish release notes, but its internal inference wire format is not a supported integration contract. A backend-shape change could therefore break—or worse, silently distort—trajectories without an obvious failure.

**Where we’d like to go instead:** Google’s current answer for headless and API-key use is the official SDK. In [[Antigravity CLI #78](https://github.com/google-antigravity/antigravity-cli/issues/78#issuecomment-4835844909)](https://github.com/google-antigravity/antigravity-cli/issues/78#issuecomment-4835844909), a project collaborator confirms that API-key authentication is not supported by `agy`, recommends the SDK instead, and says it should fit CI workflows. The corresponding request for a configurable `agy` model endpoint and custom authentication remains open in [[#514](https://github.com/google-antigravity/antigravity-cli/issues/514)](https://github.com/[google-antigravity](https://github.com/google-antigravity/antigravity-sdk-python)/antigravity-cli/issues/514).

The [`google-antigravity`[ SDK](https://github.com/google-antigravity/antigravity-sdk-python)](https://github.com/google-antigravity/antigravity-sdk-python) bundles the same underlying `localharness` engine and exposes supported endpoint overrides. That removes the proxy, CA, and real credential from the sandbox.

The honest deltas are at a minimum:

- The CLI’s automatic `AGENTS.md`/`GEMINI.md` loading is absent on the SDK path, but we can inject those files ourselves.
- Google-backed `search_web` is ineffective without Google authentication; `read_url_content` still works for ordinary unauthenticated pages, subject to sandbox egress.
- Browser automation requires Chrome, which neither implementation currently provisions.
- The SDK and CLI share the underlying harness and core tool loop, but we have not established byte-for-byte prompt and configuration parity.

So: is the SDK path viable for your use case, or is there a CLI-specific behaviour that pushed you toward `agy` itself? If the SDK works, we’d welcome a PR along those lines and would be glad to review it closely.

Appreciate the effort here—it is the right problem, and we’d like to land a safe version with you.

sjawhar:
Closing: superseded. We extracted the Antigravity CLI into a standalone plugin package (inspect-antigravity) rather than landing it inside inspect_swe, so this no longer reflects our intended direction. Thanks for the look. (Posted by Claude on Sami's behalf.)

sjawhar:
(Claude, on Sami's behalf.) Correcting the record on this PR, since my earlier closing note is now wrong.

I closed this saying we had extracted the Antigravity CLI into a standalone `inspect-antigravity` plugin package "rather than landing it inside inspect_swe". We subsequently reversed that decision and landed Antigravity inside inspect_swe after all, as a native SDK agent, in #94 — which is open and is what we now carry.

So the accurate status is: this PR's `agy`-via-mitmproxy approach is abandoned (your objection to it stands and we agreed with it), but the direction stated in my closing comment is not what we did. #94 supersedes this, not a standalone package.

Apologies for the confusing trail — flagging it because you reviewed here and are reviewing our other inspect_swe PRs, and the two records contradicted each other.

PR #88, opened 2026-07-20
feat(codex_cli): configure sandbox and approval policies

sjawhar:
## Summary
- Add Codex CLI sandbox mode and workspace-write network access controls, with an explicit opt-in for pre-approving static MCP servers.
- Resolve one effective sandbox mode AND approval policy from arguments plus validated `config_overrides`, used consistently for command construction and bridged-MCP configuration.
- Fail fast on the two configurations that previously broke silently: a headless prompting `approval_policy` (codex exec hard-overrides it to `never`), and a restricted `sandbox_mode` on an image with no `bwrap`.

## Behavior
- `sandbox_mode` applies only to model-generated shell commands. Restricted modes require a system `bwrap` in the sandbox image (the codex release archive bundles none) — construction probes and raises with an actionable message instead of codex panicking mid-eval. `read-only` blocks writes and network with no approval path under the default policy; `workspace-write` remounts `.git`/`.codex` read-only in writable roots.
- `approval_policy` non-`"never"` values raise under headless exec unless an approvals reviewer is configured (`codex exec` discards `-c approval_policy` — verified on releases 0.42.0–0.145.0); they work as documented in centaur and auto-review setups.
- `network_access` is an explicit workspace-write option, default `True` (note: inverts codex's own default to preserve continuity with the previous full-access behavior); the `config_overrides` escape hatch wins via later-`-c` precedence.
- Static `mcp_servers` keep Codex's per-server approval gate by default; `approve_static_mcp_tools=True` opts them into the same pre-approval as bridged servers. **Default-path behavior note:** bridged servers now receive `default_tools_approval_mode="approve"` under the default `approval_policy="never"` — previously the gate never fired only because `danger-full-access` auto-approved everything.
- `config_overrides["sandbox_mode"]` is now intercepted and validated like `approval_policy`, so requesting a restricted mode via overrides can no longer emit the full-access bypass flag.

## Coordination
Rebased onto main including the merged #91; this PR is Codex CLI only and no longer overlaps it.

Written by Claude, acting for Sami (@sjawhar).


src/inspect_swe/_codex_cli/config.py
```diff
@@ -51,3 +51,25 @@ def codex_cli_config_overrides(
...
+    MCP tool calls have their OWN approval gate
+    (`default_tools_approval_mode`, one of "prompt"/"writes"/"auto"/"approve"),
+    separate from the top-level `approval_policy`: with `approval_policy="never"`
+    alone, write-type MCP tool calls (e.g. an `edit_file` call) are auto-denied
+    ("user cancelled MCP tool call") rather than run, because headless
+    `codex exec` has no way to answer the resulting approval prompt. "auto" is
+    NOT sufficient either (confirmed empirically: same auto-denial) -- only
+    "approve" actually skips the gate. When `approval_policy` is `"never"`, set
+    the per-server gate to `"approve"` so it doesn't silently override the
+    caller's intent.
```
craig-meridian: This last clause reads backwards — the function *does* override the per-server gate; the point is that it only does so when the caller has already asked for `never`. Something like: "Only applied when `approval_policy` is `\"never\"`, so a caller who chose a prompting policy keeps the per-server default."

The rest of the docstring is excellent — the empirical finding that `"auto"` is insufficient and only `"approve"` skips the gate is exactly the kind of thing that gets re-discovered painfully later. For a reader's benefit you could cite the enum: `AppToolApproval` in `codex-rs/config/src/mcp_types.rs:21-27` (`auto`/`prompt`/`writes`/`approve`, defaulting to `Auto`).

src/inspect_swe/_claude_code/claude_code.py
```diff
@@ -71,7 +71,7 @@ def claude_code(
     haiku_model: str | None = None,
     subagent_model: str | None = None,
     filter: GenerateFilter | None = None,
-    auto_mode: bool = False,
+    permission_mode: str | None = None,
```
craig-meridian: **[P1] Preserve the released `auto_mode` API.**

`auto_mode` has been part of the public `claude_code()` signature since 0.2.56 (`git tag --contains d3f72ce`), and `claude_code` is exported from `inspect_swe/__init__.py`. Agent kwargs are routinely set from task configs and `-S` args, so replacing the parameter outright means `claude_code(auto_mode=True)` raises `TypeError` before an evaluation starts — with no migration path.

This repo already has an established convention for exactly this case: `**deprecated_args: Unpack[CodexDeprecatedArgs]` plus a `resolve_*_deprecated_args()` helper (`_codex_cli/config.py:8-27`, wired in at `codex_cli.py:94`). Suggest applying the same pattern here — map `auto_mode=True` to `permission_mode="auto"`, and raise if both are supplied with conflicting values.

There's a second reason beyond back-compat: #91 is open against the same base, and its `allowlist_mcp_tools` docstring tells callers to "Set `False` with `auto_mode=True`". Whichever of the two lands second, that instruction dangles. Keeping `auto_mode` as an alias means #91 needs no change here at all.

src/inspect_swe/_claude_code/claude_code.py
```diff
@@ -121,7 +121,12 @@ def claude_code(
         haiku_model: The model to use for haiku, or [background functionality](https://code.claude.com/docs/en/costs#background-token-usage). Defaults to `model`.
         subagent_model: The model to use for [subagents](https://code.claude.com/docs/en/sub-agents). Defaults to `model`.
         filter: Filter for intercepting bridged model requests.
-        auto_mode: Use `auto` permission mode rather than `--dangerously-skip-permissions`. Note that this can result in rejected tool calls so only enable if your evaluation can tolerate this.
+        permission_mode: Run under this Claude Code `--permission-mode`
+            (e.g. `"acceptEdits"`, `"auto"`, `"plan"`) instead of
+            `--dangerously-skip-permissions`. Note this can result in rejected
+            tool calls so only set if your evaluation can tolerate this; bridged
+            MCP tools still need `--allowed-tools` to be usable at all under any
+            non-bypass mode (handled automatically -- see `resolve_mcp_servers`).
```
craig-meridian: **[P2] `permission_mode` should be a `Literal`, not a bare `str`.**

The CLI exposes a closed set — from `claude --help` on 2.1.181:

```
--permission-mode <mode>  (choices: "acceptEdits", "auto", "bypassPermissions", "default", "dontAsk", "plan")
```

As typed, `permission_mode="acceptedits"` fails deep inside the sandbox rather than at type-check time, and every other option in this signature is precisely typed. A `Literal` of those six would also let the docstring drop the "e.g." hedging.

Two things worth stating outright in the docstring, since these modes aren't interchangeable in ways that matter here:

- `bypassPermissions` (near-equivalent to the default `--dangerously-skip-permissions` path) and `dontAsk` are now reachable — that's what a caller wanting "no gating, but not the scary flag" will be looking for.
- `auto` is the mode that engages Claude Code's first-party classifier, and it's the only one the allow-list interaction in #91 concerns. Listing it beside `acceptEdits` and `plan` as an undifferentiated example undersells that.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -85,6 +86,8 @@ def codex_cli(
     env: dict[str, str] | None = None,
     user: str | None = None,
     sandbox: str | None = None,
+    sandbox_mode: Literal["read-only", "workspace-write", "danger-full-access"] = "danger-full-access",
+    approval_policy: Literal["untrusted", "on-request", "never"] = "never",
```
craig-meridian: **[P1] `ruff format --check` fails on this file** — line 89 exceeds the formatter's width and needs wrapping (the other spot is the over-wrapped `cmd.extend` at 297-299). `make check` fixes both.

Two suggestions while you're here:

- Hoist these two unions into named aliases in `config.py` next to `CodexWebSearch` (e.g. `CodexSandboxMode`, `CodexApprovalPolicy`) — that also gives `codex_mcp_server_toml` a precise parameter type instead of `str`.
- Nit: `sandbox_mode` sits directly beside the existing `sandbox` parameter, which is an Inspect sandbox *environment name* — easy to conflate. A cross-reference in the docstring ("not related to the `sandbox` option above") would save a reader the double-take.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -125,6 +128,20 @@ def codex_cli(
...
         sandbox: Optional sandbox environment name.
+        sandbox_mode: Codex's own sandbox policy for model-generated shell commands
+            (`-s`/`--sandbox`). Defaults to `"danger-full-access"`, which combined with
+            `approval_policy="never"` (the default) reproduces the original
+            `--dangerously-bypass-approvals-and-sandbox` behavior. Passing
+            `"workspace-write"` or `"read-only"` runs Codex's own Linux sandbox
+            (bubblewrap); this requires the sandbox environment to allow unprivileged
+            user namespace creation, and `sandbox_workspace_write.network_access` is
+            forced on so bridged MCP tools and the model proxy (both localhost HTTP)
+            keep working.
```
craig-meridian: **[P2] The stated rationale for forcing `network_access` is incorrect.**

Codex's own flag help is explicit that the sandbox policy applies "when executing model-generated **shell commands**" (`codex-rs/utils/cli/src/shared_options.rs:37-40`). MCP servers and the model proxy are contacted by the unsandboxed Codex parent process, not from inside a sandboxed shell — so bridged tools and the proxy keep working regardless of `network_access`.

What the override actually does is re-enable network for the *agent's own shell commands*, which is a substantial part of what a caller selecting `workspace-write` was trying to prevent. Please correct the reasoning here; see the note on the code below for the behavior itself.

Also worth a sentence: with the default `approval_policy="never"`, `sandbox_mode="read-only"` means writes are *denied outright* with no approval path, i.e. "the agent cannot edit anything." That's a legitimate configuration, but a reader of "runs Codex's own sandbox" may reasonably expect prompts instead.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -263,10 +280,24 @@ async def codex_config_toml() -> str:
...
+            # via config override (`codex exec` has no `-a`/`--ask-for-approval` flag).
+            if sandbox_mode == "danger-full-access" and approval_policy == "never":
+                cmd.append("--dangerously-bypass-approvals-and-sandbox")
+            else:
+                cmd.extend(["--sandbox", sandbox_mode])
+                cmd.extend(["-c", f"approval_policy={approval_policy}"])
+                if sandbox_mode == "workspace-write":
+                    cmd.extend(
+                        ["-c", "sandbox_workspace_write.network_access=true"]
+                    )
```
craig-meridian: **[P2] Silently re-enabling network defeats much of the point of `workspace-write`.**

Given the rationale doesn't hold (see the docstring note above), this deserves to be the caller's choice rather than an unconditional default — either a `network_access: bool = True` parameter, or at minimum a docstring line stating that `config_overrides={"sandbox_workspace_write.network_access": "false"}` turns it back off.

That escape hatch does work, for what it's worth: `build_cli_overrides_layer` (`codex-rs/config/src/overrides.rs:7-13`) applies `-c` pairs in order into one layer, so the later `config_overrides` emission wins over this one. Worth an explicit test if you intend to rely on it.

**[P2] And this whole branch is untested.** It's the riskiest new logic in the PR and has no coverage, because it's inline in the closure. Extracting it as a pure helper in `config.py` — where `codex_cli_config_overrides` and friends already live — would make the three cases directly assertable:

```python
def codex_sandbox_args(
    sandbox_mode: CodexSandboxMode, approval_policy: CodexApprovalPolicy
) -> list[str]:
    ...
```

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -263,10 +280,24 @@ async def codex_config_toml() -> str:
...
+
             # apply config overrides
             if config_overrides:
                 for key, value in config_overrides.items():
```
craig-meridian: **[P2] Use the *effective* approval policy, not the raw argument.**

`config_overrides` is emitted after the explicit `-c approval_policy=…` above, and Codex applies `-c` pairs in order into a single layer (`build_cli_overrides_layer`, `codex-rs/config/src/overrides.rs:7-13`), so a caller override wins at the CLI level. But the MCP TOML at line 325 is still generated from the stale `approval_policy` function argument.

Concretely: `approval_policy="on-request"` together with `config_overrides={"approval_policy": "never"}` runs Codex with `never` while leaving MCP tools in prompting mode — reintroducing exactly the headless auto-denial this PR sets out to fix, and doing it silently. The reverse pairing produces the mirror-image inconsistency.

Suggest resolving one effective policy up front and using it for both command construction and `codex_mcp_server_toml` — or rejecting `approval_policy` (and arguably `sandbox_workspace_write.*`) in `config_overrides` with a clear error.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -291,8 +322,11 @@ async def codex_config_toml() -> str:
             if all_mcp_servers:
                 for mcp_server in all_mcp_servers:
                     toml_config[f"mcp_servers.{mcp_server.name}"] = (
-                        mcp_server.model_dump(
-                            exclude={"name", "tools"}, exclude_none=True
+                        codex_mcp_server_toml(
+                            mcp_server.model_dump(
+                                exclude={"name", "tools"}, exclude_none=True
+                            ),
+                            approval_policy,
```
craig-meridian: **[P1] Don't auto-approve caller-supplied MCP servers.**

`all_mcp_servers` is `list(mcp_servers or []) + bridge.mcp_server_configs` (line 321), so `default_tools_approval_mode="approve"` is applied to static, caller-configured servers as well as bridged ones. In a deliberately restricted `workspace-write` or `read-only` run, that silently pre-approves every tool call from arbitrary remote/static MCP servers, including destructive ones — and Codex's own semantics for `never` are to *reject* actions requiring approval, so this inverts the caller's expressed intent for those servers.

The requirement in the PR description only concerns bridged tools. Suggest scoping the override to `bridge.mcp_server_configs` and passing static servers through unchanged, or gating static servers behind an explicit opt-in.

(Note this also silently changes the *default* path, since `approval_policy="never"` is the default — worth calling out in the PR body.)

Worth flagging that this is a pattern rather than a one-off: `claude_code.py:242` builds the same `list(mcp_servers or []) + bridge.mcp_server_configs` and then treats the result uniformly, and #91 runs into the identical distinction from the opposite direction (there, `allowlist_mcp_tools=False` strips pre-approval from *bridged* tools, where a classifier rejection breaks the eval rather than protecting it). Bridge-generated servers are the eval author's own host-side code; static ones are third-party. If it's cheap to introduce that distinction once, both agents and both PRs benefit.

src/inspect_swe/_codex_cli/config.py
```diff
@@ -51,3 +51,25 @@ def codex_cli_config_overrides(
         "web_search": f'"{web_search}"',
         "features.goals": "true" if goals else "false",
     }
+
+
+def codex_mcp_server_toml(
+    mcp_server_dump: dict[str, Any], approval_policy: str
```
craig-meridian: Type this as the shared `CodexApprovalPolicy` alias rather than `str`, so an invalid policy is caught at check time instead of silently taking the `!= "never"` branch. The existing tests loop over plain strings (`("untrusted", "on-request")`), which stays valid under a `Literal`.

src/inspect_swe/_codex_cli/config.py
craig-meridian: This last clause reads backwards — the function *does* override the per-server gate; the point is that it only does so when the caller has already asked for `never`. Something like: "Only applied when `approval_policy` is `\"never\"`, so a caller who chose a prompting policy keeps the per-server default."

The rest of the docstring is excellent — the empirical finding that `"auto"` is insufficient and only `"approve"` skips the gate is exactly the kind of thing that gets re-discovered painfully later. For a reader's benefit you could cite the enum: `AppToolApproval` in `codex-rs/config/src/mcp_types.rs:21-27` (`auto`/`prompt`/`writes`/`approve`, defaulting to `Auto`).

tests/test_codex_config.py
```diff
@@ -1,4 +1,11 @@
...
+from inspect_swe._codex_cli.config import (
+    codex_cli_config_overrides,
+    codex_config_options,
+    codex_mcp_server_toml,
+    resolve_codex_deprecated_args,
+    resolve_codex_web_search,
+)
 from inspect_swe._codex_cli.config import (
     codex_cli_config_overrides,
     codex_config_options,
```
craig-meridian: **[P1] Duplicate import block — this fails required CI.**

The new import block was added above without removing this one, so the file imports from `_codex_cli.config` twice:

```
tests/test_codex_config.py:1:1: I001 Import block is un-sorted or un-formatted
tests/test_codex_config.py:10:5: F811 Redefinition of unused `codex_cli_config_overrides` from line 3
tests/test_codex_config.py:11:5: F811 Redefinition of unused `codex_config_options` from line 4
tests/test_codex_config.py:12:5: F811 Redefinition of unused `resolve_codex_deprecated_args` from line 6
tests/test_codex_config.py:13:5: F811 Redefinition of unused `resolve_codex_web_search` from line 7
```

Delete lines 9-14; the block above is a superset. `ruff check --fix` handles it.

tests/test_claude_code_mcp_allowed_tools.py
```diff
@@ -0,0 +1,35 @@
...
+consults the allow-list (e.g. ``--permission-mode auto`` -- as opposed to
+``--dangerously-skip-permissions``, which bypasses the allow-list check
+entirely) every bridged MCP tool call is denied. Confirmed empirically
+against the real ``claude`` CLI (2.1.205 and 2.1.212) in a fresh sandbox
+home: the single-underscore rule is ignored and tool calls report "you
+haven't granted it yet"; the double-underscore rule is honored and the same
+tool call succeeds.
+"""
+
+from inspect_ai.tool import MCPServerConfig
```
craig-meridian, line 11: `ruff check` flags `I001` here (import block un-sorted/un-formatted). The fix is to drop the blank line between the two imports — ruff treats `inspect_swe` as third-party in this config (no `known-first-party`, and `src = ["."]` doesn't pick up `src/inspect_swe`), which is why `tests/test_codex_config.py` has them adjacent. I confirmed this by running the project's ruff over both variants in isolation.

Which brings up a bigger problem with this file: **#91 adds a file at this exact path**, so the two conflict add/add (`git merge-tree` on the two heads reports `CONFLICT (add/add): tests/test_claude_code_mcp_allowed_tools.py`, plus a content conflict in `claude_code.py`). #91's version already has the import grouping right and its `test_default_registers_and_allowlists_explicit_mcp_tools` is near-identical to `test_explicit_tool_list_is_unaffected` here — same `taiga-mcp` fixture, same two tool names. Worth deciding which PR owns this module and having the other add to it.

Separately: this module docstring is a model of how to document a non-obvious fix, and it's independently confirmable — the `claude` binary itself constructs `mcp__${s.serverName}__*`, so the double-underscore form is the CLI's own canonical rule. (Anthropic also documents bare `mcp__<server>` as "all tools from this server"; `mcp__<server>__*` is fine and better-scoped, just worth knowing both exist if you hit a version honoring only one.)

One consequence of the fix that neither PR states, and that I think is the crux of how these two interact: because the single-underscore rule was silently ignored, `tools="all"` servers — the default, since `MCPServerConfig.tools` defaults to `"all"` — were *never* pre-approved. Making the wildcard work flips them under `--permission-mode auto` from "not pre-approved" to "fully pre-approved, classifier bypassed." That makes #91's `allowlist_mcp_tools=False` knob *more* necessary after this lands, not less, and it's a behavior change worth a line in the PR body.

craig-meridian (changes requested):
Thanks for this and nice work.

A few things to address before this lands; specifics are inline. The blocking ones are the failing Ruff checks, the unshimmed removal of `auto_mode` (public API since 0.2.56), and static MCP servers being auto-approved alongside bridged ones.

Two items with no natural inline home:

- **Body/title vs. diff** — the PR title (which will form the release notes) indicate this is just related to Codex CLI, but I think this applies to others too.
- **Relationship to #91** — how are these two meant to fit together? They're both open against the same base and don't merge (content conflict in `claude_code.py`, add/add on `tests/test_claude_code_mcp_allowed_tools.py`,  which both PRs create), and #91 documents its new option in terms of the  `auto_mode` this PR removes. Happy to sequence them however you prefer — just want to make sure I'm reviewing them in the right order.

The PR review was heavily AI assisted (though I did review all the findings) - feel free to push back if you disagree on any of the points and I'll take a closer look.

sjawhar:
Claude on Sami's behalf.

I split the work by component. #88 is now Codex CLI only; #91 owns all Claude Code changes, including the allowed-tools wildcard test. Both branches are independently based on current upstream `main`, so they no longer conflict or need sequencing.

Addressed here:
- Fixed the Ruff formatting/import failures and added direct tests for the extracted sandbox-argument helper.
- Added typed `CodexSandboxMode` and `CodexApprovalPolicy` aliases.
- Kept static caller-supplied MCP server configuration unchanged. Only Inspect-bridged servers receive the headless `approval_policy="never"` approval override.
- Made workspace-write network access explicit through `network_access: bool = True`; it is not silently forced.
- Resolved one effective approval policy before command and bridged-MCP configuration, including validated `config_overrides["approval_policy"]`.
- Corrected the sandbox documentation: it applies to model-generated shell commands only; the parent process reaches the model proxy and MCP servers outside that sandbox. It also now states that read-only plus the default `never` policy denies writes outright.

I did not push back on the review points.

Verification: `uv run make check` passed; `uv run pytest -q --ignore=tests/test_k8s.py` passed with 112 passed and 82 skipped.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -85,6 +90,9 @@ def codex_cli(
     env: dict[str, str] | None = None,
     user: str | None = None,
     sandbox: str | None = None,
+    sandbox_mode: CodexSandboxMode = "danger-full-access",
+    approval_policy: CodexApprovalPolicy = "never",
+    network_access: bool = True,
```
craig-meridian: Nit: these three land mid-signature, ahead of `version`, `config_overrides`, `debug` and the rest. `codex_cli` has no `*` marker, so every parameter is positional-or-keyword and this shifts the position of everything after it.

Realistically nobody calls a 25-parameter agent constructor positionally, so I don't feel strongly — but the just-merged #91 settled this convention for `claude_code` by placing `allowlist_mcp_tools` at the end of the signature, so the same should apply here. (A third option: add `*` to make everything after `name` keyword-only, which retires the question permanently — breaking in its own right, so probably its own PR.)

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -125,6 +133,25 @@ def codex_cli(
...
+            an Inspect sandbox environment. Defaults to `"danger-full-access"`, which
+            combined with `approval_policy="never"` (the default) reproduces the
+            original `--dangerously-bypass-approvals-and-sandbox` behavior. Passing
+            `"workspace-write"` or `"read-only"` runs Codex's own Linux sandbox
+            (bubblewrap); this requires the sandbox environment to allow unprivileged
+            user namespace creation. The policy applies only to model-generated shell
+            commands: the Codex parent process contacts MCP servers and the model proxy
+            outside that sandbox. With the default `approval_policy="never"`, a
+            `"read-only"` sandbox denies writes outright because no approval path is
+            available.
```
craig-meridian: **Blocking, I think:** the prerequisite is understated in a way that makes both new sandbox modes unusable on a stock sandbox image — and the framing is only accurate for recent codex versions.

The docstring names unprivileged userns creation, but the harder requirement is that a `bwrap` binary exists at all. The release tarball this library downloads contains exactly one file (the `codex` binary — verified for 0.145.0), and `_util/tarball.py:15` asserts single-member extraction, so the bundled `codex-resources/bwrap` fallback can never exist in an inspect_swe install. With no system `bwrap` on PATH either, Codex doesn't degrade — it aborts:

```rust
// codex-rs/linux-sandbox/src/launcher.rs:42-47
BubblewrapLauncher::Unavailable => panic!(
    "bubblewrap is unavailable: no system bwrap was found on PATH and no bundled \
     codex-resources/bwrap binary was found next to the Codex executable")
```

**Failure scenario.** A plain Inspect docker sandbox (`python:3.12-bookworm` — no `bubblewrap` package) with `sandbox_mode="workspace-write"` → every model-generated shell command dies at sandbox launch, and with the default `approval_policy="never"` there's no escalation path, so the agent can execute nothing. Even with bwrap installed, Docker's default seccomp profile denies unprivileged user-namespace creation — `clone` is allowed only when the namespace-flag mask (including `CLONE_NEWUSER`) is zero, `unshare` requires `CAP_SYS_ADMIN`, `clone3` returns ENOSYS, default action EPERM (verified in `moby/profiles/seccomp/default_linux.go`; the Docker docs table row about `CLONE_NEWUSER` describes the kernel's capability gating, not the profile). Host policy stacks further restrictions on top — e.g. Ubuntu 24.04's AppArmor userns confinement. Codex's own warning strings anticipate exactly this ("setting up uid map: Permission denied", `codex-rs/sandboxing/src/bwrap.rs:30-35`).

The framing is also version-dependent, which matters given `version=` supports pins and `"sandbox"`: at 0.98.0 bwrap was still opt-in (hidden `--use-bwrap-sandbox`, default false) and ≤0.77.0 the sandbox was Landlock+seccomp — no bwrap, no userns, needs kernel Landlock instead. So "bubblewrap … unprivileged user namespace creation" is only true for recent releases.

Options: probe for `bwrap` when a non-full-access mode is requested and fail fast with an actionable message; document the image prerequisite prominently; or ship the standalone `bwrap-<arch>.tar.gz` release artifact alongside the binary — though note that last one is a real project, not a tweak: the `agentbinary.py` machinery resolves, caches, and installs exactly one executable, so it needs multi-asset support plus placement of `codex-resources/bwrap` beside the codex binary. Any of these beats a panic mid-eval; the fail-fast probe is probably the best effort-to-value.

Two docstring additions while you're here, both confirmed upstream:

- Under `workspace-write`, writable roots are cwd + `/tmp` + `$TMPDIR`, and top-level `.git` (plus `.agents`/`.codex`) inside each writable root is re-mounted **read-only** (`default_read_only_subpaths_for_writable_root`, `codex-rs/protocol/src/permissions.rs`). Model-run `git commit`/`checkout`/`stash` fail — a sharp edge for exactly the SWE-style evals likely to reach for this mode. (Pleasant flip side worth noting: this makes the `<agent_cwd>/.codex` CODEX_HOME tamper-proof against the model — a nice eval-integrity property.)
- `read-only` also blocks *network* for model-generated commands, not just writes (`ReadOnly { network_access: false }` default) — "denies writes outright" undersells it.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -125,6 +133,25 @@ def codex_cli(
...
+        approval_policy: Codex's approval policy (`AskForApproval`). Defaults to
+            `"never"` (commands are never escalated to interactive approval --
+            required for headless `codex exec`, which cannot answer approval
+            prompts). `-a`/`--ask-for-approval` is not accepted by `codex exec`, so
+            this is applied via `-c approval_policy=<value>`.
```
craig-meridian: **Blocking:** `codex exec` discards `-c approval_policy`, so this parameter is inert on the headless path — and setting it to anything other than `"never"` actively breaks bridged tools.

`exec` hard-sets the policy as a harness override:

```rust
// codex-rs/exec/src/lib.rs:422-427
let overrides = ConfigOverrides {
    // Default to never ask for approvals in headless mode.
    approval_policy: Some(AskForApproval::Never),
```

and config resolution gives the harness override precedence over the config layer that `-c` populates (`codex-rs/core/src/config/mod.rs:3602`: `approval_policy_override.or(cfg.approval_policy)`); `build_exec_config` only relents when `approvals_reviewer == AutoReview`. This isn't just source-reading: running real release binaries, `codex exec --sandbox workspace-write -c approval_policy=on-request` prints `approval: never` in the session banner on **0.42.0, 0.50.0, 0.63.0, 0.77.0, 0.98.0 and 0.145.0** — every version tried across ten months of releases. Upstream's own test (`exec/tests/suite/approval_policy.rs`) has to set `approvals_reviewer = "auto_review"` alongside for `on-request` to survive. Writing the key to `config.toml` instead wouldn't help — same layer, same precedence.

**Failure scenario.** `codex_cli(sandbox_mode="read-only", approval_policy="on-request")` with bridged tools: the runtime policy is still `Never`, so approvals behave no differently — but `codex_mcp_server_toml` sees a non-`never` policy and omits `default_tools_approval_mode = "approve"`. The bridged server falls back to `AppToolApproval::Auto`; under a restricted profile the full-disk-write auto-approve shortcut doesn't fire; `requires_mcp_tool_approval(None)` → `destructive_hint.unwrap_or(true)` → true. Headless, nobody answers, and every bridged Inspect tool call returns "user cancelled MCP tool call" — the exact failure this PR exists to fix. The caller asked for an approval path, got none, and lost their tools.

Two qualifications for precision. Headless, the override survives only when `approvals_reviewer = "auto_review"` is also configured (that's the `build_exec_config` escape hatch) — so "ignored" describes the normal headless path, and auto-review is a supported exception a fix shouldn't foreclose. And in centaur mode the parameter works as documented: the TUI leaves the harness override unset unless `-a` is given (`codex-rs/tui/src/lib.rs:915-924`), so `-c approval_policy` flows through there.

Suggested fix: raise (or at least warn) when `centaur is False`, `approval_policy != "never"`, and no `approvals_reviewer` is configured via `config_overrides`; and key the bridged-server `approve` override on the policy Codex will *actually run* — i.e. apply it whenever the effective runtime policy is `Never` (headless without auto-review), regardless of what the caller passed. Don't make it unconditional, though: in centaur and auto-review setups a prompting policy is real, and force-approving bridged tools there would bypass an approval flow the caller deliberately chose. Document the parameter as effective in centaur/TUI and auto-review setups only.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -125,6 +133,25 @@ def codex_cli(
...
+        network_access: Whether Codex's `"workspace-write"` sandbox may access the
+            network. Defaults to `True`; ignored by other sandbox modes.
```
craig-meridian: Worth noting here that this default inverts Codex's own: `sandbox_workspace_write.network_access` defaults to `false` upstream (`codex-rs/core/config.schema.json`, `/definitions/SandboxWorkspaceWrite/properties/network_access`).

`True` is the right default for this library in my view — it preserves continuity for anyone moving off the previous unconditional full-access behavior, where the network was always reachable. Just worth one clause so a reader who knows Codex's config doesn't assume parity. (For avoidance of doubt, the localhost model proxy and bridged MCP servers are safe either way — those connections come from the codex parent process outside the command sandbox; verified empirically under both restricted modes. `network_access=false` only gates model-generated commands, which additionally get `CODEX_SANDBOX_NETWORK_DISABLED=1`.)

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -153,6 +180,9 @@ def codex_cli(
         cast(dict[str, Any], deprecated_args)
     )
     effective_web_search = resolve_codex_web_search(web_search, disallowed_tools)
+    effective_approval_policy = resolve_codex_approval_policy(
+        approval_policy, config_overrides
```
craig-meridian: Sitting directly below `resolve_codex_web_search`, this highlights a gap: that function validates its argument at runtime (`if web_search not in (...): raise ValueError`), as does `resolve_codex_deprecated_args`, precisely because agent kwargs arrive from task configs and `-S` args as arbitrary strings where a `Literal` gives no protection. `sandbox_mode` and `approval_policy` reach the CLI the same way but are checked only by mypy.

Odd consequence inside `resolve_codex_approval_policy`: it validates the value from `config_overrides` but not the `approval_policy` argument itself. So `codex_cli(approval_policy="on_request")` (underscore) passes construction and surfaces as a clap error mid-eval, while the identical typo via `config_overrides` raises immediately.

`network_access` has the same gap: an arbitrary value reaches `str(network_access).lower()` unchecked, so `network_access="nope"` from a task config emits `-c sandbox_workspace_write.network_access=nope` and fails only when Codex parses it mid-evaluation.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -263,14 +293,20 @@ async def codex_config_toml() -> str:
...
+                codex_sandbox_args(
+                    sandbox_mode, effective_approval_policy, network_access
+                )
+            )
+
             # apply config overrides
             if config_overrides:
                 for key, value in config_overrides.items():
-                    cmd.extend(["-c", f"{key}={value}"])
+                    if key != "approval_policy":
```
craig-meridian: This filter is the right call, but it makes the `config_overrides` docstring inaccurate — line 160-161 still promises "Each key-value pair is passed as `-c key=value` to the CLI", and `approval_policy` is now deliberately consumed instead. Worth a sentence saying it's intercepted and validated (and that invalid values raise), since silently swallowing a key someone explicitly passed is otherwise surprising.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -286,15 +322,17 @@ async def codex_config_toml() -> str:
...
+            # Register static MCP servers unchanged. Bridged tools are supplied by the
+            # evaluation author, so headless approval can safely skip their MCP gate.
+            for mcp_server in mcp_servers or []:
+                toml_config[f"mcp_servers.{mcp_server.name}"] = mcp_server.model_dump(
+                    exclude={"name", "tools"}, exclude_none=True
+                )
+            for mcp_server in bridge.mcp_server_configs:
+                toml_config[f"mcp_servers.{mcp_server.name}"] = codex_mcp_server_toml(
+                    mcp_server.model_dump(exclude={"name", "tools"}, exclude_none=True),
+                    effective_approval_policy,
```
craig-meridian, line 480: **Blocking:** splitting the loops was my suggestion last round, so this is me following through on the consequence rather than reopening it — static servers are now left auto-denied under the very modes this PR adds, and callers have no supported way to fix it. I'm treating it as blocking because it means two public options (`sandbox_mode` and `mcp_servers`) silently don't compose.

The comment's rationale — bridged tools are author-supplied, so skipping their gate is safe — applies just as well to `mcp_servers=[...]`, which is also author-supplied configuration, not something the model can reach for. And "callers can supply their own approval settings" isn't actually available: inspect_ai's `MCPServerConfig` is a closed pydantic model with no `default_tools_approval_mode` field, so there is no supported way to express it.

**Failure scenario.** `codex_cli(sandbox_mode="workspace-write", mcp_servers=[MCPServerConfig(name="taiga", type="http", url=...)])`. Policy is `never`, but the profile is restricted so the full-disk-write auto-approve shortcut doesn't fire; the server's mode defaults to `AppToolApproval::Auto` (`codex-rs/config/src/mcp_types.rs:20-26`); any tool without a `readOnlyHint` annotation is cancelled. The only escape is the undocumented `config_overrides={"mcp_servers.taiga.default_tools_approval_mode": '"approve"'}`.

Notably this is *not* a regression on the default path — `danger-full-access` gives unrestricted filesystem access, so full-disk-write auto-approves everything and the distinction never shows up. It only bites the new modes, which is what makes it easy to miss.

An opt-in (`approve_static_mcp_tools: bool = False`, or per-server) would cover it — and the just-merged #91 established exactly this pattern with `allowlist_mcp_tools` on `claude_code`, a boolean controlling how static MCP servers are treated under permission gating. At minimum the asymmetry deserves documenting on `sandbox_mode`, since "use a real sandbox" and "use a static MCP server" are individually reasonable choices that silently don't compose.

src/inspect_swe/_codex_cli/config.py
```diff
@@ -51,3 +53,70 @@ def codex_cli_config_overrides(
...
+    `codex exec` has no way to answer the resulting approval prompt. "auto" is
+    NOT sufficient either (confirmed empirically: same auto-denial) -- only
+    "approve" actually skips the gate. The override is applied only when
+    `approval_policy` is `"never"`, so callers who choose a prompting policy keep
+    the per-server default.
+    """
+    mcp_server_toml = dict(mcp_server_dump)
+    if approval_policy == "never":
+        mcp_server_toml["default_tools_approval_mode"] = "approve"
+    return mcp_server_toml
```
craig-meridian, line 193: One version-compatibility caveat worth capturing here, since `version="auto"` uses whatever codex is baked into the sandbox image. The MCP tool-approval gate this override defeats originally applied only to the built-in Codex Apps server — at 0.98 (and still at 0.116) `maybe_request_mcp_tool_approval` early-returns for any other server (`if server != CODEX_APPS_MCP_SERVER_NAME { return None; }`), so arbitrary configured servers bypassed it entirely. That early return is gone at 0.117, extending the gate to all MCP servers, while the `default_tools_approval_mode` key it's answered with only landed in v0.122 (#17843). So the vulnerable window is **v0.117–v0.121**: the gate applies, the key is **silently ignored** (confirmed: older versions parse a config.toml containing it without error), and restricted-sandbox runs with bridged tools hit the auto-denial this function exists to prevent, with no signal that the fix didn't take.

Good news on the other side: the key is silently ignored rather than a parse error back to at least 0.50 (`RawMcpServerConfig` uses `#[schemars(deny_unknown_fields)]`, not serde's), so there's no config-parse regression on old pins.

Worth a documented version floor (≥0.122 for the MCP override to function) — or a runtime check if the binary version is already known at setup time.

One docstring correction while you're in here: "with `approval_policy="never"` alone, write-type MCP tool calls … are auto-denied" is only true under a **restricted sandbox**. Under full disk access (the default `danger-full-access` path) `never` auto-approves MCP writes without this override (`mcp_permission_prompt_is_auto_approved`: `Approve` mode, or `Never` + full-disk-write). So the override is redundant on the default path and load-bearing only for the new modes — worth saying, since it explains where the empirical denial was observed.

src/inspect_swe/_codex_cli/config.py
```diff
@@ -51,3 +53,70 @@ def codex_cli_config_overrides(
...
+                "-c",
+                f"sandbox_workspace_write.network_access={str(network_access).lower()}",
+            ]
+        )
+    return sandbox_args
+
+
+def resolve_codex_approval_policy(
+    approval_policy: CodexApprovalPolicy,
+    config_overrides: Mapping[str, str] | None,
```
craig-meridian, line 255: Three things about this function.

**1. It rejects values upstream accepts.** `"on-failure"` is a valid policy string (serde alias of `OnRequest`, `codex-rs/protocol/src/protocol.rs:917`, and a genuinely distinct variant on older pinned versions) and now raises at construction where it previously passed through. (For completeness: upstream also has a `granular` policy, but it's a data-carrying variant — `Granular(GranularApprovalConfig)`, `protocol.rs:927` — expressible only as a TOML table, so rejecting the bare string `"granular"` is correct; only the `"on-failure"` rejection is a real regression.)

**2. It rejects TOML-quoted values, which worked before.** `-c key=value` is parsed as TOML by Codex with a raw-string fallback, so `approval_policy="never"` and `approval_policy=never` are both valid — and this repo's own `codex_cli_config_overrides` emits the quoted form (`f'"{web_search}"'`), so quoting is the local convention a caller would copy. `codex_cli(config_overrides={"approval_policy": '"never"'})` worked on `main` and now raises. Stripping surrounding quotes before the match would fix that case.

**3. `sandbox_mode` gets no equivalent treatment, and that fails in the unsafe direction — I'd treat this one as blocking.** `codex_cli(config_overrides={"sandbox_mode": "read-only"})` with otherwise-default arguments still emits `--dangerously-bypass-approvals-and-sandbox`, and that flag reaches Codex separately from the config-override layer (`ExecRunArgs.dangerously_bypass_approvals_and_sandbox`, `codex-rs/exec/src/lib.rs:210`) so it wins. The caller believes they asked for a read-only sandbox and gets full access, silently — the inverse of the failure this function was added to prevent, and a one-line fix (reject the key) now that the setting is a first-class parameter.

Cheapest fix is probably to reject `sandbox_mode` in `config_overrides` outright now that it's a first-class parameter, rather than resolving it. Same argument for `sandbox_workspace_write.network_access` now that `network_access` exists. (To be precise about the current state: for both of those keys a user override *does* win, just via different mechanisms — `approval_policy` by being intercepted, validated, and resolved into the effective policy; `network_access` by being emitted after the generated value, where Codex's last-value-wins applies. The inconsistency is in validation and handling — one path validates and one doesn't — not in who wins. `sandbox_mode` is the only key where the override silently loses.)

tests/test_codex_config.py
```diff
@@ -51,3 +56,76 @@ def test_codex_cli_config_overrides_format_values_for_cli() -> None:
...
+        assert codex_mcp_server_toml(dump, policy) == dump
+
+
+def test_codex_mcp_server_toml_does_not_mutate_input() -> None:
+    dump = {"type": "http", "url": "http://localhost:8901/mcp/taiga-mcp"}
+    codex_mcp_server_toml(dump, "never")
+    assert "default_tools_approval_mode" not in dump
+
+
+@pytest.mark.parametrize(
```
craig-meridian: Nice — this is exactly the coverage the inline version couldn't have. A few gaps:

- `codex_sandbox_args` covers 3 of 9 mode×policy combinations. The load-bearing misses are `("danger-full-access", "on-request")` — the case that abandons the bypass flag while still naming the default mode — and `("workspace-write", "never", True)`, which is the actual default network path.
- `network_access=True` never reaches an assertion, so nothing pins the `=true` spelling. Since the value goes through `str(network_access).lower()`, a regression to Python's `True` capitalisation passes today.
- `resolve_codex_approval_policy` is untested for `config_overrides=None`, `{}`, and a dict without the key — the `None` branch only gets exercised via a call that *does* carry the key.
- Nothing pins the accepted-value set against upstream, which would have caught the `"on-failure"`/`"granular"` divergence noted on `config.py`.

Bigger picture: nothing in `tests/` inspects the constructed Codex command line at all, so every integration point this PR adds is unverified — that `codex_sandbox_args` output actually reaches `cmd` and in the right position, that the `if key != "approval_policy"` filter behaves, and that bridged server tables get `default_tools_approval_mode` while static ones don't. A single test asserting the assembled `cmd` and `toml_config` for two or three configurations would cover all of that wiring. (To be clear about its limits: it would *not* catch the headless `approval_policy` interaction discussed above — the assembled command containing `-c approval_policy=on-request` looks perfectly correct; that one needs a real-binary integration test, or an explicit compatibility test encoding the upstream behavior.)

craig-meridian (changes requested):
Thanks for the rework — this addresses every point from the last round and I've resolved those threads. The following review was written by Claude with my supervision.

A deeper pass, now that the work has been split between #91 and this PR (including running real codex release binaries against the flag/config shapes this PR generates) turned up problems I missed the first time. Four I'd consider blocking:

1. `codex exec` discards `-c approval_policy` — confirmed empirically on six releases from 0.42.0 to 0.145.0 — so the parameter is headless-inert, and non-`never` values actively break bridged tools.
2. The new sandbox modes need a `bwrap` binary that never reaches the sandbox, so on a stock image every shell command dies at sandbox launch.
3. Static `mcp_servers` are silently unusable under the new restricted modes, with no supported way to fix it — `sandbox_mode` and `mcp_servers` don't compose.
4. `config_overrides={"sandbox_mode": "read-only"}` still emits the full-access bypass flag — a caller who asked for a restricted sandbox silently gets none.

There's also a version-compatibility window (0.117–0.121) where the MCP approval fix is silently ignored. Details inline; the rest are smaller consistency/documentation follow-ups that could equally become issues.

sjawhar:
Claude here, on Sami's behalf. Before changing anything we independently reproduced all four blockers, and every one held up:

1. Real release binaries 0.98.0 and 0.145.0 both print `approval: never` for `exec --sandbox workspace-write -c approval_policy=on-request`.
2. The 0.145.0 tarball contains exactly one member, and `codex sandbox /bin/true` in a stock `python:3.12-bookworm` container panics with the exact launcher message. (We also hit the misleading "will use the bundled bubblewrap" warning on a path that then panics.)
3. Confirmed at source level: `MCPServerConfig` exposes no approval field, `AppToolApproval` defaults to `Auto`, and the managed-profile auto-approve check requires full-disk write.
4. Confirmed from our own generated args: the bypass flag was emitted alongside `-c sandbox_mode=read-only`.

Addressed in the latest push (rebased onto main including #91, so the sequencing question is moot):

- **Effective values everywhere:** `config_overrides["sandbox_mode"]` is now intercepted and validated like `approval_policy`; command construction and the bridged-MCP TOML both derive from the single effective policy/mode. Regression test asserts the override can never co-emit the bypass flag.
- **Headless guard:** non-`never` `approval_policy` without an approvals reviewer now raises at construction with the centaur/auto-review paths spelled out, rather than silently cancelling every bridged tool call. Auto-review setups are not foreclosed: `config_overrides={"approvals_reviewer": ...}` suppresses the guard.
- **bwrap fail-fast:** restricted modes probe for `bwrap` in the sandbox before launch and raise with an actionable message (your ranked-best option); the docstring now states the single-binary-tarball reality, the Docker seccomp/userns caveat, and the version dependence (Landlock ≤0.77, opt-in at 0.98).
- **Static servers:** pass through unchanged by default; new `approve_static_mcp_tools: bool = False` opts them into the bridged treatment, with the composition asymmetry documented on both options. The default-path behavior change for bridged servers is now called out in the PR body.
- Runtime validation for `sandbox_mode`/`approval_policy`/`network_access` arguments (matching the existing `web_search` convention), the `network_access` default-inversion note and escape hatch documented, the 0.117–0.121 version window captured on `codex_mcp_server_toml`, new parameters moved to the end of the signature per the #91 convention, and the `config_overrides` docstring documents the interception.

`make check` and the full test suite pass (129 passed).

sjawhar:
One follow-up to the push above: main's newly-landed auto_review feature manages the same controls, so the two are now integrated explicitly — auto_review supplies the effective workspace-write/on-request values (its final -c overrides keep winning at the CLI level, and no --sandbox/bypass flag is emitted alongside), the bwrap fail-fast probe covers auto_review runs too, and explicitly combining auto_review with sandbox_mode/approval_policy (or those config_overrides keys) raises rather than silently fighting over precedence. Full suite passes (154).

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -316,15 +459,26 @@ async def codex_config_toml() -> str:
...
+            # evaluation author, so headless approval always skips their MCP gate;
+            # static caller-configured servers keep Codex's per-server default
+            # unless approve_static_mcp_tools opts them in (under the restricted
+            # sandbox modes that default cancels un-annotated tool calls
+            # headlessly -- see the approve_static_mcp_tools docstring).
+            for mcp_server in mcp_servers or []:
+                static_toml = mcp_server.model_dump(
+                    exclude={"name", "tools"}, exclude_none=True
+                )
+                if approve_static_mcp_tools:
```
craig-meridian, line 472: **[P1] Validate the new static-MCP opt-in as a real boolean (new since round two).**

`approve_static_mcp_tools` was added in response to the second review, so this code was not present in either earlier pass. Unlike `network_access`, it receives no runtime validation. Agent kwargs can arrive as strings, so `approve_static_mcp_tools="false"` is truthy and enters this branch, silently pre-approving static MCP tools under the default `never` policy. In a restricted mode that reverses the caller's explicit opt-out. Please require an actual `bool` during construction and add a regression test for the string `"false"`.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -289,16 +423,25 @@ async def codex_config_toml() -> str:
...
                 ]
             )
-            # with auto_review, approvals/sandbox come from config (on-request +
-            # workspace-write); the bypass flag would force approval_policy=never
-            # at a precedence -c can't beat
+            # Sandbox/approval args from the effective values. With auto_review,
+            # approvals/sandbox come from its final -c overrides instead
+            # (on-request + workspace-write): an explicit --sandbox or bypass
+            # flag would force settings at a precedence -c can't beat.
             if resolved_auto_review is None:
```
craig-meridian, line 430: **[P2] Honor or explicitly disallow `network_access` with `auto_review` (new interaction since round two).**

`auto_review` landed on `main` after the second review, and this integration was added in the final merge commit, so the interaction did not exist in the revisions previously reviewed. With `auto_review` enabled the effective mode is `workspace-write`, but this branch skips `codex_sandbox_args`, so the validated `network_access` argument is never emitted. Even the documented/default `network_access=True` therefore leaves Codex's upstream default of `false` in effect. Either thread the setting into the final auto-review overrides, or explicitly reject/document the combination if guardian-gated network is intentional.

craig-meridian (changes requested):
Thanks, Sami, for the thoughtful and thorough rework here. The major round-two blockers are addressed, and both local checks and CI pass. I have two remaining inline comments, both concerning code or interactions added since that review: the new `approve_static_mcp_tools` opt-in and the integration with the newly merged `auto_review` feature. Details are inline.

src/inspect_swe/_codex_cli/codex_cli.py
sjawhar: Claude here, on Sami's behalf. Implemented in b27ab9fa: `approve_static_mcp_tools` is validated as a real `bool` at construction, with a regression test for the string `"false"`.

src/inspect_swe/_codex_cli/codex_cli.py
sjawhar: Claude here, on Sami's behalf. Implemented in b27ab9fa: `network_access` is now threaded into the auto_review path, with tests covering both values.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -227,6 +348,27 @@ async def execute(state: AgentState) -> AgentState:
             # resolve sandbox
             sbox = sandbox_env(sandbox)
 
+            # Codex's restricted sandbox modes shell out to bubblewrap, and the
+            # codex release archive ships no bundled bwrap (single-binary
+            # tarball), so a system bwrap must exist in the sandbox image.
+            # Without one codex PANICS at sandbox launch on every
+            # model-generated shell command -- fail fast with an actionable
+            # error instead of a mid-eval panic.
+            if effective_sandbox_mode != "danger-full-access":
```
craig-meridian: **[P1] The probe fires for every non-full-access *effective* mode — which includes `auto_review`, breaking a documented, currently-working configuration.**

`docs/codex_cli.qmd:66` (landed with #102/#103) explicitly supports `auto_review` on bwrap-less images — which it calls "the default for most images": each sandboxed command fails, a capable model re-issues with `require_escalated`, and the guardian adjudicates — review is model-driven rather than OS-enforced. After this PR, `codex_cli(auto_review=True)` on such an image raises `RuntimeError` at startup instead, and the error's advice (`use sandbox_mode='danger-full-access'`) is unusable for those users, since combining `auto_review` with an explicit `sandbox_mode` now raises.

The probe is also unconditional on version and platform, where the docstring above it is properly qualified: codex <= 0.77 sandboxes via Landlock+seccomp (no bwrap), 0.98 is bwrap-opt-in, and macOS uses Seatbelt. So `codex_cli(version="0.77.0", sandbox_mode="read-only")` — or `version="sandbox"` with an older installed binary — now fails preflight even though that binary never launches bwrap.

Suggested fix: gate the probe on the already-resolved `codex_version` (and platform) actually using bwrap, and exempt `auto_review` — or downgrade to a warning there — so the documented model-driven fallback keeps working. If you'd rather make bwrap a hard prerequisite for `auto_review` too, that's a defensible position, but then it's a behavior change to a shipped feature and the docs' "OS sandbox unavailable" section needs rewriting in the same PR.

Aside, while you're in this block: the probe re-runs on every `execute()` of the agent (multi-call solvers, resume paths), re-verifying an immutable image fact with a blocking sandbox round-trip. Caching the result in `store()` (the `MODEL_PORT` pattern above) or folding it into the adjacent `resolve_agent_cwd` exec avoids that.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -165,6 +233,59 @@ def codex_cli(
...
+        )
+
+        # Headless `codex exec` hard-overrides the runtime approval policy to
+        # `never` (harness override, precedence over `-c`; verified empirically
+        # on 0.42.0-0.145.0), EXCEPT when an approvals reviewer is configured. A
+        # prompting policy here would therefore not prompt -- but it WOULD
+        # disable the bridged-MCP approval override below, cancelling every
+        # bridged tool call. Fail fast instead of silently losing the agent's
+        # tools.
+        if (
```
craig-meridian, line 310: **[P1] The escape hatch this guard recommends isn't version-guarded, so following the error message's own advice can reproduce the exact failure the guard exists to prevent.**

The `codex exec` relent that honors `approvals_reviewer` headlessly (`build_exec_config` rebuilding with `approval_policy: None` when the resolved reviewer is `AutoReview`) first appears at codex **0.137.0** — verified against codex-rs at `rust-v0.137.0` vs `rust-v0.136.0-alpha.2`; every earlier release unconditionally hard-overrides the policy to `never`, and before ~0.116 the key is an ignored unknown. This repo already encodes that floor (`CODEX_AUTO_REVIEW_MIN_VERSION = "0.137.0"`, whose docstring says exactly this), but `check_codex_auto_review_version` is only invoked when `resolved_auto_review is not None`.

Failure: `codex_cli(approval_policy="on-request", config_overrides={"approvals_reviewer": '"auto_review"'})` with `version="auto"` resolving a pre-0.137 image binary → runtime policy is still `never`, while the effective policy tracked here is `"on-request"` — so `codex_mcp_server_toml` omits `default_tools_approval_mode="approve"` and every bridged write tool call is cancelled silently: the precise outcome this `ValueError` was added to prevent, reached via the path its message recommends.

Fix looks like one line: run `check_codex_auto_review_version(codex_version)` on the reviewer-escape-hatch path too (it already handles `None` versions gracefully).

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -316,15 +469,26 @@ async def codex_config_toml() -> str:
...
+            # evaluation author, so headless approval always skips their MCP gate;
+            # static caller-configured servers keep Codex's per-server default
+            # unless approve_static_mcp_tools opts them in (under the restricted
+            # sandbox modes that default cancels un-annotated tool calls
+            # headlessly -- see the approve_static_mcp_tools docstring).
+            for mcp_server in mcp_servers or []:
+                static_toml = mcp_server.model_dump(
+                    exclude={"name", "tools"}, exclude_none=True
+                )
+                if approve_static_mcp_tools:
```
craig-meridian: **[P3] Optional bundle — three smaller items, all in code this PR adds. None block approval; take any you want, and an explicit pass on the rest is fine (say so and I'll resolve this thread).**

1. **`approve_static_mcp_tools=True` is silently inert whenever the effective policy isn't `"never"`** — notably `codex_cli(auto_review=True, mcp_servers=[...], approve_static_mcp_tools=True)` changes nothing (effective policy is `on-request`, so `codex_mcp_server_toml` is a no-op), while the sibling controls (`sandbox_mode`/`approval_policy`) fail loudly in the same combination. Either add it to the `auto_review` guard or note the interaction on both docstrings.
2. **Non-bool `centaur` values bypass the new headless guard.** The guard keys off `centaur is False` identity, and `-S` values arrive via `yaml.safe_load` with no annotation-based coercion (verified against the installed inspect_ai): `centaur=0`/`None`/quoted `"false"` all skip the ValueError and then produce a malformed headless run (`exec` omitted, prompting policy, no reviewer) — the same arbitrary-kwarg class this PR just fixed for `network_access`/`approve_static_mcp_tools`. Normalizing/validating `centaur` alongside those closes it. (Bare `-S centaur=false` does coerce; quoted values, config files, and programmatic callers don't.)
3. **The static-server gating branch here is untested.** `test_static_mcp_server_toml_opt_in_path` asserts helper behavior that `test_codex_mcp_server_toml_sets_approve_when_never` and `..._leaves_other_policies_untouched` already pin, and never exercises this loop — a regression here (gating on the wrong flag, or applying the helper to the wrong server class) passes the suite. One test asserting the assembled `toml_config` for a static+bridged configuration covers the branch and most of the wiring the round-two test thread asked about.

src/inspect_swe/_codex_cli/config.py
```diff
@@ -157,3 +157,162 @@ def check_codex_auto_review_version(version: str | None) -> None:
...
+) -> list[str]:
+    """CLI args selecting Codex's own sandbox and approval policy.
+
+    With `auto_review` the mode and policy come from its own final `-c`
+    overrides (`workspace-write` + `on-request`) instead: an explicit
+    `--sandbox` or bypass flag would pin them at a precedence `-c` cannot beat.
+    Its effective mode is still `workspace-write`, so network access is
+    configured on that path too.
+    """
+    if auto_review:
```
craig-meridian: **[P3] Optional nit:** `codex_sandbox_args` now takes `sandbox_mode` and `approval_policy` that it entirely ignores when `auto_review=True`. Since the call site already knows `resolved_auto_review`, choosing there — `codex_network_access_args(network_access) if resolved_auto_review else codex_sandbox_args(...)` — deletes the flag parameter, the ignored-arguments semantics, and `test_codex_sandbox_args_auto_review_never_emits_mode_or_policy`, which exists only to pin that oddity. Fine to decline; just say so.

src/inspect_swe/_codex_cli/config.py
craig-meridian: Following up so this thread doesn't dangle into another round: point 3 (the `sandbox_mode` bypass) was fixed — thanks. Points 1–2 are still the case on the current head, reproduced at b27ab9f: `config_overrides={"approval_policy": '"never"'}` and `{"sandbox_mode": '"read-only"'}` (the TOML-quoted forms `main` accepted, and the quoting convention `codex_cli_config_overrides` itself uses) now raise `ValueError` at construction, as does the valid upstream `"on-failure"` spelling.

Happy with either resolution: **fix** (strip surrounding quotes before matching; accept `"on-failure"`, mapping it to `on-request` for the effective-policy logic), or an explicit **"intentionally stricter now — won't do"**, in which case I'll resolve this thread as accepted behavior. Either answer is fine; I just don't want existing eval configs failing on upgrade to be an accident rather than a decision.

src/inspect_swe/_codex_cli/config.py
craig-meridian: The version-window half of this comment was addressed in the docstring — thanks. The other half is still open: the docstring's "with `approval_policy=\"never\"` alone, write-type MCP tool calls … are auto-denied" remains unqualified, but that's only true under a **restricted sandbox** — under the default `danger-full-access` profile, `never` auto-approves MCP writes via the full-disk-write shortcut, so this override is redundant on the default path and load-bearing only for the new modes. One sentence scoping it closes this; equally fine to decline explicitly and I'll resolve the thread.

craig-meridian (changes requested):
Thanks — both round-three items are fixed exactly as asked ([P1] boolean validation, [P2] `network_access` under `auto_review`); I verified the fixes including command-line ordering, and `make check` plus the full test suite pass locally on this head. The following review was again written by Claude with my supervision.

This round I ran the resulting behavior against the shipped docs and codex release history, which surfaced two blockers in code from the earlier rework that I should have caught then — they're late, and I own that, but they're not scope creep: both would break real configurations on merge.

**To approve, I need:**

1. The two [P1]s below (bwrap probe vs. `auto_review`/older versions; the `approvals_reviewer` escape hatch missing its version floor).
2. `docs/codex_cli.qmd` brought up to date. It isn't touched by this PR, but this PR changes what it documents: the options table lists every option *except* the four this PR adds; "**Network**: unavailable by default" (line 78) and the guardian-adjudicating-network-escalations language (lines 54/66) are inverted now that `network_access=True` applies under `auto_review`; the line-80 example recommends the exact override that has become the default (the meaningful example is now the `"false"` direction or the first-class option); and the `config_overrides` row should mention that `approval_policy`/`sandbox_mode` are intercepted and validated.

Everything else is explicitly optional, marked [P3]: take whatever you want; declining is completely fine.

One process ask so I'm not left guessing: the two still-open round-two threads (quoted/`on-failure` override values; the MCP auto-denial docstring qualifier) each need an explicit answer — either a fix, or a clear "won't do" stated on the thread. Please don't leave them unanswered: if a reply says "won't do", I'll resolve the thread as a deliberate decision, but an unanswered thread I have to treat as still open. The same applies to any [P3] here you choose to decline.

With the two [P1] fixes and the docs page, I expect this to be the final round.

craig-meridian:
Verified the latest pushes: both [P1]s are fixed exactly as asked — the probe gating even took the caching aside, nice — the docs page is thoroughly updated, CI is green, and the full suite passes locally. The MCP-docstring thread is addressed too, and I've resolved it along with everything else the pushes have covered, so the thread list now reflects only what's actually outstanding.

The one thing between this and approval is the process ask from my review: the three threads still open each need an explicit answer — a fix, or a stated "won't do". No further code is required for approval; a one-line reply per thread does it, and I'll pick this up on my next pass once they're there.

src/inspect_swe/_codex_cli/config.py
sjawhar: Claude here, on Sami's behalf. Points 1 and 2 are fixed in d80719afad82.

`resolve_codex_approval_policy` and `resolve_codex_sandbox_mode` now strip one layer of surrounding double quotes before validating, so the TOML-quoted forms work again — matching the convention `codex_cli_config_overrides` itself emits. `approval_policy` also accepts `on-failure`, normalised to `on-request`.

On the mapping: I checked codex-rs rather than take it on faith — `AskForApproval` carries `#[serde(alias = "on-failure")] OnRequest` in `protocol/src/protocol.rs`, and there's an upstream unit test asserting that exact deserialization, so it resolves to the same variant rather than merely a similar one. Your note about `granular` holds and it still raises: it's a data-carrying variant expressible only as a table, so a bare string shouldn't satisfy it, quoted or not.

Reproduced all three failures at `b27ab9f` first, then added regression tests for quoted `never`, quoted `read-only`, and `on-failure` in both quoted and unquoted forms, alongside the existing unquoted coverage. Invalid values still raise.

src/inspect_swe/_codex_cli/codex_cli.py
sjawhar: Claude here, on Sami's behalf. Took all three, in d80719afad82 — each was smaller to fix than to argue for declining.

1. `approve_static_mcp_tools=True` now raises when combined with `auto_review=True` instead of being silently inert, so it fails loudly exactly like `sandbox_mode` and `approval_policy` already do in that combination. The bool validation moved above the guard so the check sees a real bool.
2. `centaur` now raises for a value that is neither a bool nor `CentaurOptions`, instead of bypassing the headless approval-policy guard.
3. Extracted the static-versus-bridged MCP-toml gating into a pure `codex_mcp_servers_toml()` helper and added a direct unit test covering both the `approve_static_mcp_tools` opt-in and the always-approved bridged path. That follows the `codex_sandbox_args` precedent from the earlier round rather than mocking the async sandbox pipeline to reach an inline closure.

src/inspect_swe/_codex_cli/config.py
sjawhar: Claude here, on Sami's behalf. Took this one as specified, in d80719afad82.

`codex_sandbox_args` no longer accepts `sandbox_mode`/`approval_policy` for the `auto_review` case — the `codex_cli` call site now chooses directly between `codex_network_access_args` and `codex_sandbox_args` using `resolved_auto_review`. The parameter is gone, along with the two tests that existed only to pin the old ignored-arguments behaviour.

craig-meridian (approved):
Thanks Sami, I'm happy to approve this and merge it.

PR #90, opened 2026-07-20
ci: static hatch-vcs fallback_version for tag-less git-URL installs

sjawhar:
## What
Set a static hatch-vcs `fallback_version` so installs from a tag-less / shallow git URL resolve a valid version instead of failing.

## Why
Installing directly from a git commit (no tags in a shallow clone) makes hatch-vcs fail to derive a version.

## Notes
Normal tagged releases are unaffected. The fallback is currently a fixed value; happy to adjust the policy if you'd prefer it track differently.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,10 @@ version-file = "src/inspect_swe/_version.py"
 
 [tool.hatch.version.raw-options]
 local_scheme = "no-local-version" # Important for PyPI compatibility
+# Used only when no tag is reachable (e.g. a shallow/tag-less `uv` git-URL
+# fetch of a single commit) -- normal tagged checkouts/releases still get a
+# real git-describe-derived version and ignore this.
+fallback_version = "0.2.66"
 
 [tool.hatch.metadata]
 allow-direct-references = true


dragonstyle (changes requested):
I think we should avoid hardcoding a version that will go stale unless it is a sentinel value. Two suggested fixes:

1) We use release please in this repo - update the release please config + add annotation comment in the file so that release please will automatically maintain that version as we release.

2) Use a sentinel 0.0.0 version if version semantics aren't important.

I'm guessing you're affecting by this, so open to your feedback on which approach better matches your needs.

sjawhar:
Closing — dropping the hardcoded `fallback_version` approach entirely rather than maintaining a version (via release-please or a sentinel).

Our fork already carries the upstream version tags (all 72, through `0.2.66`), so hatch-vcs resolves a real git-describe version on tag-ful git-URL installs without any fallback — e.g. `release/2026-07-24.2` describes as `0.2.66-11-g<sha>` and a downstream install resolves `0.2.67.devN`. That removes the need for a static/sentinel version here. Thanks for the review and the release-please/sentinel options.

PR #91, opened 2026-07-21
feat(claude_code): control static MCP allowlisting

sjawhar:
## Summary
- Add `allowlist_mcp_tools` for static caller-provided MCP servers.
- Preserve the released `auto_mode` API through the established deprecated-argument shim while exposing the complete typed Claude Code permission-mode set.
- Correct the MCP wildcard syntax to `mcp__<name>__*` and cover the default `tools="all"` path.
- Keep Inspect-bridged MCP tools pre-approved when static MCP tools are sent to Claude Code's classifier.

## Behavior
`allowlist_mcp_tools=False` applies only to static `mcp_servers`. Bridged tools remain allowlisted because evaluation-owned submit and scoring tools can be required for the task to complete. In `permission_mode="auto"`, static MCP calls that are not pre-approved are adjudicated by Claude Code's first-party classifier.

Existing `tools="all"` static MCP servers were previously emitted as `mcp__<name>_*`, which Claude Code does not recognize as an MCP wildcard. They are now emitted as `mcp__<name>__*`, so existing `auto_mode=True` and `permission_mode="auto"` callers that rely on the default `allowlist_mcp_tools=True` now pre-approve every tool from those servers instead of sending each call to the classifier.

## Coordination
This PR is Claude Code only. #88 independently owns Codex CLI sandbox and approval-policy controls. Both branches are based on upstream `main` and do not depend on each other.

## Release notes

feat(claude_code): add `permission_mode` option (deprecates `auto_mode`)

fix(claude_code): correct "all" tools MCP allowlist wildcard to `mcp__<name>__*`

### DIFF
--- a/src/inspect_swe/_claude_code/claude_code.py
+++ b/src/inspect_swe/_claude_code/claude_code.py
@@ -2,7 +2,7 @@ import shlex
 import uuid
 from pathlib import Path
 from textwrap import dedent
-from typing import Any, Literal, Sequence
+from typing import Any, Literal, Mapping, Sequence, cast
 
 from inspect_ai.agent import (
     Agent,
@@ -28,6 +28,7 @@ from inspect_ai.util import (
 )
 from pydantic import Field
 from pydantic_core import to_json
+from typing_extensions import TypedDict, Unpack
 
 from inspect_swe._claude_code._events.live_consumer import LiveConsumer
 from inspect_swe._claude_code._events.stream import (
@@ -47,6 +48,55 @@ from .._util.sandbox import resolve_agent_cwd
 from .._util.trace import trace
 from .agentbinary import claude_code_binary_source
 from .model import resolve_claude_code_models
+
+ClaudeCodePermissionMode = Literal[
+    "acceptEdits", "auto", "bypassPermissions", "default", "dontAsk", "plan"
+]
+
+
+class ClaudeCodeDeprecatedArgs(TypedDict, total=False):
+    auto_mode: bool
+
+
+def resolve_claude_code_deprecated_args(
+    deprecated_args: Mapping[str, Any],
+    permission_mode: str | None,
+) -> ClaudeCodePermissionMode | None:
+    unexpected_args = set(deprecated_args) - {"auto_mode"}
+    if unexpected_args:
+        unexpected = ", ".join(sorted(unexpected_args))
+        raise TypeError(f"Unexpected keyword argument(s): {unexpected}")
+
+    auto_mode = deprecated_args.get("auto_mode")
+    if auto_mode:
+        if permission_mode is not None and permission_mode != "auto":
+            raise ValueError(
+                "auto_mode=True conflicts with permission_mode; use permission_mode='auto'."
+            )
+        return "auto"
+    return resolve_claude_code_permission_mode(permission_mode)
+
+
+def resolve_claude_code_permission_mode(
+    permission_mode: str | None,
+) -> ClaudeCodePermissionMode | None:
+    match permission_mode:
+        case None:
+            return None
+        case (
+            "acceptEdits"
+            | "auto"
+            | "bypassPermissions"
+            | "default"
+            | "dontAsk"
+            | "plan"
+        ):
+            return permission_mode
+        case _:
+            raise ValueError(
+                "permission_mode must be one of 'acceptEdits', 'auto', "
+                "'bypassPermissions', 'default', 'dontAsk', or 'plan'."
+            )
 
 
 @agent
@@ -71,7 +121,7 @@ def claude_code(
     haiku_model: str | None = None,
     subagent_model: str | None = None,
     filter: GenerateFilter | None = None,
-    auto_mode: bool = False,
+    permission_mode: ClaudeCodePermissionMode | None = None,
     retry_refusals: int | None = 3,
     retry_uncaught_errors: int | None = 3,
     cwd: str | None = None,
@@ -80,6 +130,8 @@ def claude_code(
     sandbox: str | None = None,
     version: Literal["auto", "sandbox", "stable", "latest"] | str = "auto",
     debug: bool | None = None,
+    allowlist_mcp_tools: bool = True,
+    **deprecated_args: Unpack[ClaudeCodeDeprecatedArgs],
 ) -> Agent:
     """Claude Code agent.
 
@@ -121,7 +173,15 @@ def claude_code(
         haiku_model: The model to use for haiku, or [background functionality](https://code.claude.com/docs/en/costs#background-token-usage). Defaults to `model`.
         subagent_model: The model to use for [subagents](https://code.claude.com/docs/en/sub-agents). Defaults to `model`.
         filter: Filter for intercepting bridged model requests.
-        auto_mode: Use `auto` permission mode rather than `--dangerously-skip-permissions`. Note that this can result in rejected tool calls so only enable if your evaluation can tolerate this.
+        permission_mode: Claude Code `--permission-mode`. The complete CLI set is
+            `"acceptEdits"`, `"auto"`, `"bypassPermissions"`, `"default"`,
+            `"dontAsk"`, and `"plan"`. `"bypassPermissions"` is near-equivalent
+            to the default `--dangerously-skip-permissions` path, and `"dontAsk"`
+            is also now reachable. `--allowed-tools` is consulted in every mode
+            except `"bypassPermissions"`. In unattended runs, tools excluded from
+            `--allowed-tools` cannot be prompted for and are denied; `"auto"` is
+            the only mode where otherwise-unapproved calls are adjudicated by
+            Claude Code's first-party classifier rather than denied unconditionally.
         retry_refusals: Should refusals be retried? Defaults to retrying up to 3 times.
         retry_uncaught_errors: Should uncaught errors (unexpected crashes of Claude Code) be retried. Defaults to retrying up to 3 times.
         cwd: Working directory to run claude code within.
@@ -135,6 +195,15 @@ def claude_code(
             - "latest": Download and use the very latest version of claude code.
             - "x.x.x": Download and use a specific version of claude code.
         debug: Add `--debug` cli flag and trace all debug output.
+        allowlist_mcp_tools: Whether to add static caller-provided MCP tools to
+            `--allowed-tools` (default `True`). It matters in every permission
+            mode except `"bypassPermissions"`: in unattended runs, excluded
+            static tools are denied without prompting. Set `False` with
+            `permission_mode="auto"` when Claude Code's first-party classifier
+            should adjudicate those tools. Bridged Inspect tools remain allowlisted
+            because an evaluation may depend on them being callable.
+        **deprecated_args: Supports the deprecated `auto_mode` argument. Set
+            `auto_mode=True` maps to `permission_mode="auto"`.
     """
     # resolve centaur
     if centaur is True:
@@ -145,6 +214,10 @@ def claude_code(
 
     # resolve attempts
     attempts = AgentAttempts(attempts) if isinstance(attempts, int) else attempts
+
+    effective_permission_mode = resolve_claude_code_deprecated_args(
+        cast(dict[str, Any], deprecated_args), permission_mode
+    )
 
     # allocate session_id once per agent instance so that all calls to execute()
     # for the same sample share the same session. this enables --resume <id> to
@@ -211,11 +284,9 @@ def claude_code(
                 claude_code_binary_source(), version, user, sandbox_env(sandbox)
             )
 
-            # base options — auto_mode uses --permission-mode auto (monitor active);
-            # otherwise --dangerously-skip-permissions (no permission gating).
             permission_flag = (
-                ["--permission-mode", "auto"]
-                if auto_mode
+                ["--permission-mode", effective_permission_mode]
+                if effective_permission_mode is not None
                 else ["--dangerously-skip-permissions"]
             )
             cmd = [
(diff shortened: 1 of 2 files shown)


src/inspect_swe/_claude_code/claude_code.py
```diff
@@ -72,6 +72,7 @@ def claude_code(
     subagent_model: str | None = None,
     filter: GenerateFilter | None = None,
     auto_mode: bool = False,
+    allowlist_mcp_tools: bool = True,
```
craig-meridian: I appreciate this may be the most logical ordering for this parameter, but given it forms part of the public API, we should put this at the end of the parameter list to not break callers relying on positional arguments.

src/inspect_swe/_claude_code/claude_code.py
```diff
@@ -235,7 +242,7 @@ async def execute(state: AgentState) -> AgentState:
             all_mcp_servers = list(mcp_servers or []) + bridge.mcp_server_configs
             if all_mcp_servers:
                 mcp_server_args, mcp_allowed_tools = resolve_mcp_servers(
-                    all_mcp_servers
+                    all_mcp_servers, allowlist_mcp_tools=allowlist_mcp_tools
```
craig-meridian: `all_mcp_servers` includes `bridge.mcp_server_configs`, so `allowlist_mcp_tools=False` strips pre-approval from bridged Inspect tools too, not just static `mcp_servers`. Those may have different trust profiles: a third-party MCP server is exactly what you'd want the classifier to scrutinise, but a bridged tool is the eval author's own host-side code — and possibly something the eval depends on the agent calling (e.g. a submit/scoring tool), where a classifier rejection breaks the eval rather than protects it. The docstring does say "bridged/static", so this is documented — just double-checking the uniform granularity is intended, vs. only un-allowlisting static servers.

tests/test_claude_code_mcp_allowed_tools.py
```diff
@@ -0,0 +1,29 @@
...
+    server = MCPServerConfig(
+        type="http", name="taiga-mcp", tools=["list_files", "read_file"]
+    )
+
+    mcp_config_cmds, allowed_tools = resolve_mcp_servers(
+        [server], allowlist_mcp_tools=False
+    )
+
+    assert mcp_config_cmds[0] == "--mcp-config"
+    assert allowed_tools == []
```
craig-meridian, line 43: nit (not strictly this PR's scope, but good opportunity given this new test module): The new tests only exercise the explicit-list branch of `resolve_mcp_servers`. The `tools="all"` wildcard branch — which is the default, since `MCPServerConfig.tools` defaults to `"all"` — has no coverage anywhere, so nothing currently pins the `mcp__<name>_*` allowlist format. Worth adding while you're in this file:

```py
def test_default_allowlists_wildcard_for_all_tools() -> None:
    server = MCPServerConfig(type="http", name="taiga-mcp", tools="all")

    mcp_config_cmds, allowed_tools = resolve_mcp_servers([server])

    assert mcp_config_cmds[0] == "--mcp-config"
    assert allowed_tools == ["mcp__taiga-mcp_*"]
```

craig-meridian (changes requested):
Thanks Sami, looks good. One blocking comment, others are questions/nits.

tests/test_claude_code_mcp_allowed_tools.py
craig-meridian: Ignore this one — the allowlist format I wrote is wrong. `mcp__<name>_*` is a
no-op (Claude Code only recognises `mcp__<name>__*`, double underscore), so a
test pinning my version would have pinned the bug. #88 fixes the format and
adds exactly this test, and it creates this same file — so no need to add
anything here; we just need to sort out which PR owns the module.

sjawhar:
Claude on Sami's behalf.

I split the work by component. #91 is now Claude Code only; #88 owns the Codex CLI sandbox and approval work. Both branches are independently based on current upstream `main`, so they no longer conflict or need sequencing.

Addressed here:
- Moved `allowlist_mcp_tools` to the end of the public parameter list.
- Preserved `auto_mode` through the established `**deprecated_args: Unpack[...]` shim. `auto_mode=True` maps to `permission_mode="auto"`, and conflicting explicit values raise.
- Typed `permission_mode` to Claude Code's complete closed set and documented that `bypassPermissions` and `dontAsk` are reachable, while `auto` is the classifier mode relevant to allowlisting.
- Corrected the `tools="all"` wildcard to `mcp__<name>__*` and added coverage for that default path.
- Scoped `allowlist_mcp_tools=False` to static caller-provided MCP servers. Bridged Inspect tools stay allowlisted because evaluation-owned submit or scoring tools may be required for an eval to complete.

I did not push back on the review points.

Verification: `uv run make check` passed; `uv run pytest -q --ignore=tests/test_k8s.py` passed with 111 passed and 81 skipped.

src/inspect_swe/_claude_code/claude_code.py
```diff
@@ -121,7 +151,12 @@ def claude_code(
         haiku_model: The model to use for haiku, or [background functionality](https://code.claude.com/docs/en/costs#background-token-usage). Defaults to `model`.
         subagent_model: The model to use for [subagents](https://code.claude.com/docs/en/sub-agents). Defaults to `model`.
         filter: Filter for intercepting bridged model requests.
-        auto_mode: Use `auto` permission mode rather than `--dangerously-skip-permissions`. Note that this can result in rejected tool calls so only enable if your evaluation can tolerate this.
+        permission_mode: Claude Code `--permission-mode`. The complete CLI set is
+            `"acceptEdits"`, `"auto"`, `"bypassPermissions"`, `"default"`,
+            `"dontAsk"`, and `"plan"`. `"bypassPermissions"` is near-equivalent
+            to the default `--dangerously-skip-permissions` path, and `"dontAsk"`
+            is also now reachable. `"auto"` alone invokes Claude Code's first-party
+            classifier; only that mode interacts with `allowlist_mcp_tools`.
```
craig-meridian: The claim that only `"auto"` interacts with `allowlist_mcp_tools` is inherited from my own comment on #88, but on reflection it's wrong — `--allowed-tools` is consulted in every permission mode except `bypassPermissions`. In unattended runs (`--print`, the default `centaur=False` path) there's no human to answer prompts, so under `"default"`, `"acceptEdits"`, `"plan"`, or `"dontAsk"` with `allowlist_mcp_tools=False`, static MCP calls aren't "reviewed" — they're denied outright and the eval silently loses those tools. (Centaur mode is the exception: it's interactive, so those modes can prompt instead.)

Could you rework this (and the `allowlist_mcp_tools` entry below) to say the flag matters in all non-bypass modes, and name the unattended failure mode? Something like: "In unattended runs, tools not in `--allowed-tools` are denied without prompting in every mode except `bypassPermissions`; `auto` is the only mode where otherwise-unapproved calls are adjudicated by a classifier rather than denied unconditionally."

src/inspect_swe/_claude_code/claude_code.py
```diff
@@ -509,7 +558,7 @@ def resolve_mcp_servers(
             exclude={"name", "tools"}, exclude_none=True
         )
         if mcp_server.tools == "all":
-            allowed_tools.append(f"mcp__{mcp_server.name}_*")
+            allowed_tools.append(f"mcp__{mcp_server.name}__*")
```
craig-meridian, line 618: This one-character fix is a real behavior change that the PR body doesn't state: the old `mcp__<name>_*` form violates the CLI's allow-rule grammar (globs are only valid after a literal `mcp__<server>__` prefix — the binary ships `mcp__puppeteer__*` as its own example), so `tools="all"` servers — the default, since `MCPServerConfig.tools` defaults to `"all"` — were *never* pre-approved. After this lands they flip to fully pre-approved, classifier bypassed, for existing `auto_mode=True` users (and new `permission_mode="auto"` callers) who haven't set the new flag.

Two asks (assuming I've understood correctly, and you're in agreement):
- a line in the PR body stating that flip, and
- a `fix(claude_code): correct "all" tools MCP allowlist wildcard to mcp__<name>__*` line in the body so it surfaces in the release notes as its own entry (release-please parses `<type>:` body lines). The `auto_mode` deprecation arguably deserves a body line too.

src/inspect_swe/_claude_code/claude_code.py
```diff
@@ -230,15 +274,19 @@ async def execute(state: AgentState) -> AgentState:
...
+            bridged_mcp_servers = bridge.mcp_server_configs
+            all_mcp_servers = static_mcp_servers + bridged_mcp_servers
             if all_mcp_servers:
-                mcp_server_args, mcp_allowed_tools = resolve_mcp_servers(
-                    all_mcp_servers
+                mcp_server_args, _ = resolve_mcp_servers(all_mcp_servers)
+                _, static_allowed_tools = resolve_mcp_servers(
+                    static_mcp_servers, allowlist_mcp_tools=allowlist_mcp_tools
                 )
+                _, bridged_allowed_tools = resolve_mcp_servers(bridged_mcp_servers)
```
craig-meridian: The PR's headline guarantee — bridged tools stay allowlisted when `allowlist_mcp_tools=False` — lives only in this inline split, which no unit test can reach: a regression that swaps the flag onto the bridged call (or drops `bridged_allowed_tools`) would pass the whole suite. Could you extract it into a small pure helper, e.g.

```py
def resolve_allowed_mcp_tools(
    static: Sequence[MCPServerConfig],
    bridged: Sequence[MCPServerConfig],
    allowlist_mcp_tools: bool,
) -> list[str]: ...
```

and assert the static/bridged asymmetry directly in the new test module? That also collapses the three `resolve_mcp_servers` calls (the helper's own `allowlist_mcp_tools` param becomes redundant at that point — validation still runs via the combined `--mcp-config` call).

tests/test_claude_code_mcp_allowed_tools.py
```diff
@@ -0,0 +1,56 @@
+import pytest
```
craig-meridian, line 10: The #88 incarnation of this file had a module docstring documenting *why* double-underscore is correct (the CLI itself constructs `mcp__<server>__*`; globs are only recognised after a literal `mcp__<server>__` prefix). That rationale got dropped in the move to this PR, so the evidence for a non-obvious rule now survives only in a closed PR. Could you restore it?

Worth one extra sentence there: bare `mcp__<name>` is the older, more portable "all tools from this server" spelling — relevant because `version="auto"`/`"sandbox"` can run whatever (possibly older) Claude Code is already installed. No need to change the emission; just record that both forms exist.

src/inspect_swe/_claude_code/claude_code.py
```diff
@@ -48,6 +49,33 @@
...
+    deprecated_args: Mapping[str, Any],
+    permission_mode: ClaudeCodePermissionMode | None,
+) -> ClaudeCodePermissionMode | None:
+    unexpected_args = set(deprecated_args) - {"auto_mode"}
+    if unexpected_args:
+        unexpected = ", ".join(sorted(unexpected_args))
+        raise TypeError(f"Unexpected keyword argument(s): {unexpected}")
+
+    auto_mode = deprecated_args.get("auto_mode")
+    if auto_mode is True:
```
craig-meridian: nit (optional): the old code was `if auto_mode:`, so a caller passing a truthy non-`True` (`auto_mode=1` from a task config or `-S` arg) previously got auto mode but now silently falls through to `--dangerously-skip-permissions` — the opposite safety posture, with no error. `if deprecated_args.get("auto_mode"):` preserves the old semantics; alternatively raise on non-bool.

Related hardening while you're in this function: runtime-validating the effective `permission_mode` against the Literal set (as `resolve_codex_web_search` does for its modes) would turn any stray value that arrives via an untyped path — task configs, or a positional argument landing in `auto_mode`'s old slot — into a clear construction-time error instead of a malformed `--permission-mode` invocation deep in the sandbox. Happy for both to be take-it-or-leave-it.

craig-meridian (changes requested):
Thanks Sami — this addresses everything from the last round and I've resolved those threads. The permission_mode/shim work matches what I asked for on #88.

Remaining items after a fresh review are inline comments. Again, I've leant on Claude Code for this, so do feel free to challenge something if it doesn't seem right. Not all are blockers.

src/inspect_swe/_claude_code/claude_code.py
sjawhar: Claude for Sami: Updated both docstring entries. They now state that `--allowed-tools` is consulted in every non-bypass mode, name the unattended denial path, and distinguish `auto` as the classifier path for otherwise-unapproved calls.

src/inspect_swe/_claude_code/claude_code.py
sjawhar: Claude for Sami: Updated the PR body. It now calls out that the malformed `tools="all"` wildcard had never pre-approved those servers, that `mcp__<name>__*` changes existing auto-mode callers to pre-approval, and includes the requested `fix(claude_code):` release-note entry. I also added entries for the toggle and `auto_mode` deprecation.

src/inspect_swe/_claude_code/claude_code.py
sjawhar: Claude for Sami: Extracted `resolve_allowed_mcp_tools(static, bridged, allowlist_mcp_tools)`. The command now serializes and validates the combined server list once, then resolves the static/bridged allowlists separately. The new direct regression proves that disabling the static allowlist still retains bridged `mcp__...` entries.

tests/test_claude_code_mcp_allowed_tools.py
sjawhar: Claude for Sami: Restored the module-level rationale for the double-underscore wildcard and added the portability note for the older bare `mcp__<name>` all-tools spelling. The emitted form remains `mcp__<name>__*`.

src/inspect_swe/_claude_code/claude_code.py
sjawhar: Claude for Sami: Changed the shim back to truthiness, so `auto_mode=1` preserves its prior auto-mode behavior. Added runtime parsing of the effective permission mode against the supported literal set, with a construction-time `ValueError` for stray values, plus regressions for both cases.

craig-meridian (approved):
Thanks Sami — all five points addressed cleanly. The extracted `resolve_allowed_mcp_tools` with the direct static/bridged regression test is nicer than what I sketched.

One thing I'll fix directly in the PR description rather than ask for another round: release-please only lifts `<type>:` lines that start their own blank-line-separated paragraph, so the Release notes section as currently formatted won't reach the changelog (and the extra `feat:` line would duplicate the title's entry). I'll reformat it before merge.

PR #94, opened 2026-07-24
feat(antigravity): add native Antigravity SDK agent

sjawhar:
## Summary

Adds a native `antigravity` agent that runs Google's `google-antigravity` SDK (which bundles the localharness engine) headless inside an Inspect sandbox, mirroring the structure of `gemini_cli`.

This is the SDK-based approach @craig-meridian invited in #85: model calls route through `sandbox_agent_bridge` over the SDK's native Gemini `generateContent` wire via a `GeminiAPIEndpoint` `base_url` override (no OpenAI translation), which keeps real credentials out of the sandbox. It replaces the earlier mitmproxy-CLI direction that was declined there.

## How it works

- The agent runs the Antigravity SDK's localharness engine headless in the sandbox.
- Inference is bridged: the SDK is pointed at the bridge endpoint via a `GeminiAPIEndpoint` `base_url` override, so all model calls go through `sandbox_agent_bridge` on the SDK's native Gemini wire.
- Credential boundary: only a dummy `GEMINI_API_KEY` enters the sandbox. The real key stays host-side, since inference is bridge-routed.
- The host agent is decoupled from `google.antigravity`: `RunnerPayload` is defined locally and `sdk_runner.py` is sandbox-only.
- Runtime SDK presence is handled by `agentbinary.ensure_antigravity_sdk`: skip if the SDK is already baked into the image, otherwise install it.

## Dependency

This depends at runtime on a Google bridge converter fix in inspect_ai (UKGovernmentBEIS/inspect_ai#4607): the bridge must re-role a `functionResponse` part that arrives inside a `role="model"` turn (the cloudcode/localharness dialect) into a proper tool message, otherwise the reconstructed conversation ends on a model turn and Gemini rejects the next request, stalling the agent after one tool call. With that fix in place, the agent completes full multi-step tool loops.

## Tests

- `tests/test_antigravity.py`: 7 host-safe unit tests, all passing; ruff clean.

## End-to-end proof

Native agent on a tool-calling benign smoke task, model `gemini-3.6-flash`: 3/3 samples scored 1.0, 0 errors, full multi-step task (list/read/edit/shell/summary/memory/web_fetch), and it self-corrected a tool schema error mid-run.

## Known follow-up (not blocking)

Provisioning for arbitrary no-egress sandboxes is not yet wired. `agentbinary` currently does skip-if-baked else a naive pip install (which needs egress). For general use it should adopt the existing wheel-bundle infrastructure (`ensure_agent_wheel_installed` / `download_wheels_tarball`, as `mini_swe_agent` does), adapted for a library import rather than a CLI tool. `google-antigravity` is on PyPI, so this is viable as a follow-up.

(Posted by Claude on Sami's behalf.)


sjawhar:
Pushed two fixes to this branch (`fc048eb`) after an Oracle + browser-eval review found them:

## 1. mypy strict (CI blocker)

`inspect_swe` runs `mypy --strict` over `src` and `tests`, which flagged 3 errors:
- `sdk_runner.py:9,10` — `google.antigravity` / `google.antigravity.hooks` are `import-untyped`. `google-antigravity` is a **sandbox-only** runtime dependency (the SDK runs inside the Inspect sandbox, not on the host), so it is intentionally not a project dependency and ships no `py.typed`. Fixed with a scoped `[[tool.mypy.overrides]]` (`module = ["google.antigravity.*"]`, `ignore_missing_imports = true`) rather than adding a host dependency — the host stays decoupled from `google.antigravity`.
- `tests/test_antigravity.py:40` — non-overlapping equality (`assert _SANDBOX_DUMMY_API_KEY != ""` on a non-empty literal). Changed to `assert _SANDBOX_DUMMY_API_KEY`.

`mypy --strict` now reports 0 issues; `ruff check` + `ruff format` clean; the 7 host-safe unit tests still pass.

## 2. Runtime provisioning correctness (`agentbinary.py`)

The egress-image provisioning path (`_provision_antigravity_sdk`, used only when the SDK is **not** already importable) had two bugs:
- It resolved a base python that lacked `pip` (an active venv on `PATH` with no `pip` module) and died with `No module named pip` before any install. `_base_python` now selects an interpreter that can build a venv with pip bootstrapped (`import ensurepip, venv`), preferring `/usr/bin/python3`.
- It installed with `pip install --user`, but the runner executes with `PYTHONNOUSERSITE=1`, which hides user-site installs — so even a successful install was invisible. It now creates a dedicated venv (`/var/tmp/.antigravity-sdk-venv`) and installs into it, whose own site-packages stays importable under `PYTHONNOUSERSITE=1`.
- Pinned `google-antigravity==0.1.7` (the version validated end-to-end).

### Verification honesty

The mypy/ruff/unit-test fixes are fully verified. The **provisioning path change is correct-by-construction but not exercised end-to-end**: it only runs on an egress image where the SDK is not baked in. In agent-c (the current consumer) the SDK is baked into the image and sandboxes are `network_mode: none`, so `_python_with_sdk` succeeds and provisioning is skipped — the e2e run validates the bake, not this path. A true e2e of the provisioning path needs an egress image without the SDK baked.

The fully-offline answer for arbitrary no-egress sandboxes (host-side wheel download + push via the existing `ensure_agent_wheel_installed` / `download_wheels_tarball` infra, adapted for a library import rather than a CLI entrypoint) remains the disclosed follow-up.

(Posted by Claude on Sami's behalf.)

sjawhar:
### Provisioning hardening follow-ups (next `inspect_swe` release, not blocking)

Consolidating the known follow-ups for the antigravity provisioning path, to be handled together in a future release (none are urgent; the current pin is self-consistent):

1. **Offline wheel-push provisioning** (already noted above): adopt the existing `ensure_agent_wheel_installed` / `download_wheels_tarball` infra (host-side wheel download + push), adapted for a library import rather than a CLI entrypoint, so no-egress sandboxes can provision without baking.

2. **Version-assert the baked SDK.** `agentbinary._python_with_sdk` currently only checks `import google.antigravity` (presence), not version. If a future release bumps `_SDK_VERSION` and a downstream image's baked pin isn't bumped in lockstep, the stale baked version is silently used (and on a no-egress image it couldn't re-provision anyway). Fix: when a baked SDK is found, assert `importlib.metadata.version("google-antigravity") == _SDK_VERSION` and fail loud on mismatch, turning silent version drift into a clear "bake pin out of sync" error.

Today `_SDK_VERSION == "0.1.7"` matches the consumer bake pin, so this is defense-in-depth, not a live defect.

(Posted by Claude on Sami's behalf.)

craig-meridian (changes requested):
Thanks for reworking this along the SDK/bridge lines from #85. Review below from Claude Code. Please lmk if you disagree with anything.

## Blockers

**1. Bump the `inspect_ai` floor to 0.3.250.** The bridge converter fix this agent depends on ([UKGovernmentBEIS/inspect_ai#4635](https://github.com/UKGovernmentBEIS/inspect_ai/pull/4635), which superseded #4607 and merged as `9d63ee1b`) shipped in inspect_ai 0.3.250 on 2026-07-28 (GitHub compare `0.3.250...9d63ee1b` → `behind_by: 32, ahead_by: 0`). The branch still declares `inspect_ai>=0.3.244`, where `_bridge/google_api_impl.py` handles `functionResponse` only in `_extract_user_parts` (line 402) — the `role == "model"` branch (line 310) drops it, so the reconstructed conversation ends on an assistant turn and the agent stalls after one tool call. Anyone installing at the declared floor gets a broken agent; `pyproject.toml` needs `inspect_ai>=0.3.250`.

**2. The public agent is hard-coded to one consumer environment, and `antigravity()` with default arguments always fails at execution time.** `_taiga_mcp_config` requires a bridged MCP server literally named `taiga-mcp`. But `bridge.mcp_server_configs` names come entirely from caller-supplied `BridgedToolsSpec.name` — nothing in inspect_ai produces `taiga-mcp`, and this repo's own example uses `name="secrets"` (`examples/bridged_tools/task.py:40`). Since `bridged_tools` defaults to `None`, the default configuration raises `RuntimeError` as soon as `execute()` runs, and any `BridgedToolsSpec` under a different name is silently unreachable even when the resolver is satisfied. The same literal is baked into the runner payload (`mcp_name: Literal["taiga-mcp"]`) and the policy predicate `_targets_configured_server`. The constructor likewise rejects any `user`/`cwd` other than `model` and `/home/model` — including `user=None`, the repo-wide default meaning "sandbox default user".

Two coherent resolutions; worth an explicit call rather than landing between them:

- **Generalize.** Pass all of `bridge.mcp_server_configs` through the payload as a list, add an `mcp_servers` parameter (every MCP-capable sibling has one), let the policy predicate allow any configured server name, and resolve the working directory with the existing `resolve_agent_cwd(sbox, user, cwd)` + `echo $HOME` pattern that `gemini_cli` uses.
- **Scope it explicitly as an adapter** for the intended image, documented as such — in which case it probably should not be exported via `_registry.py` and `__init__.py`'s `__all__` as a peer of `claude_code()` / `gemini_cli()`. Registry membership is what makes something a documented public agent with the API-consistency obligations that come with it.

**3. Repeated invocations silently drop conversation history.** `antigravity.py` discards `build_user_prompt`'s second return (`prompt, _ = ...`), and `sdk_runner.run()` always constructs a fresh SDK `Agent`. `build_user_prompt` only includes user messages after the last assistant message, so on a second invocation (handoff back, operator follow-up) all prior context is silently lost — no error, just an agent with amnesia. Every sibling binds `has_assistant_response` and resumes (`gemini_cli.py:163` + `--resume latest` at `:223`; same pattern in claude_code, codex_cli, opencode, mini_swe_agent). This isn't an SDK limitation: the pinned 0.1.7 exposes `Agent.conversation_id` for resuming a saved session via the config's `conversation_id` (the SDK docstring says `SessionConfig.conversation_id`, but in 0.1.7 the field lives on `LocalAgentConfig` — the exact object the runner already constructs, with `save_dir` already configured). Persist the conversation id across runs and resume — or, if that's deferred, fail loudly on re-invocation rather than silently truncating.

**4. The SDK presence probe doesn't run under the runner's import isolation.** `_python_with_sdk` execs `import google.antigravity` with no environment, but the selected interpreter is later launched with `PYTHONNOUSERSITE=1` (`sdk_execution_spec`). A user-site-only installation (`pip install --user`) therefore passes the probe, provisioning is skipped, and the real runner then dies with `ModuleNotFoundError`. The agentbinary docstring shows the failure class is understood — the venv-over-`--user` choice exists precisely because `PYTHONNOUSERSITE=1` hides user-site installs — the probe just wasn't given the same treatment. Fix is one line (`sandbox.exec(..., env={"PYTHONNOUSERSITE": "1"})` on the probe and the post-install verify), plus a regression test.

## Tool confinement: verified forced — keep the allowlist, but make drift loud

I checked the claim in the code comments against `google-antigravity==0.1.7` directly, since it's load-bearing for the design. It holds:

- `CapabilitiesConfig.enabled_tools` / `disabled_tools` have exactly the right semantics (per the SDK's own docstring, a disabled tool "is stripped from the model's context entirely"), but both are typed `list[BuiltinTools]`, a closed 13-member enum. The four engine tools this PR needs to strip (`list_resources`, `read_resource`, `manage_task`, `schedule`) are not members, and pydantic rejects them as strings with a `ValidationError`. `call_mcp_tool` likewise.
- There's no Python seam to reach around it: the declaration is assembled inside `google/antigravity/bin/localharness`, a compiled binary. All five names appear in the binary's strings and in none of the SDK's Python files.

So `enabled_tools=[]` already delegates all 13 nameable builtins to the harness, and the `GenerateFilter` covers only the residual four the SDK's type system cannot express. That's a supplement to delegation, not a replacement for it — worth saying in the code comment, since #91 (merged) and #88 delegate tool restriction to their vendor's config surface and a future reader will reasonably ask why this agent doesn't.

The allowlist polarity is right for an eval harness (fail closed on capabilities), and I found no present capability leak. The gap is that drift is *silent*: if a future localharness version renames the dispatcher or adds new engine builtins, the filter quietly drops names and the first symptom is downstream (denial loops or missing tools), with nothing in the log tying it to the SDK version. Two additions would close that, and they reinforce each other:

- In the filter, log a warning (or raise) whenever a declared tool name falls outside allowlist ∪ {the four known engine tools} — SDK drift then becomes a visible, attributable event with zero capability leak.
- The baked-SDK version assert already proposed in the follow-ups comment on this thread (item 2). Non-blocking as disclosed, but it's cheap and is the front line against the drift scenario above, so worth considering for this PR.

## Non-blocking: API and design

- **`model_aliases` is effectively inert as exposed.** The bridge resolves aliases against the *client-sent* model name (`google_api_impl.py:70`), and the client name is always the hidden literal `_BRIDGE_ENDPOINT_MODEL = "gemini-3.6-flash"` (`sdk_runner.py:13`) — so the only alias key that can ever match is an undocumented string. Carry the endpoint model through `RunnerPayload` (compare `gemini_cli`'s `gemini_model` parameter); that fixes the inert parameter and the hidden constant together.
- **Gate tracing behind `debug`.** `trace()` dumps full runner stdout+stderr unconditionally; `gemini_cli` gates the equivalent behind its `debug` parameter.
- **Expose `version`.** `_SDK_VERSION = "0.1.7"` is pinned in code with no override; `gemini_cli` and `mini_swe_agent` both expose a `version` parameter for exactly this.
- **`filter` type divergence — intentional?** The rebase re-typed `filter` from `GenerateFilter` to a local Model-first `_ModelGenerateFilter` alias, which cleanly avoids the wrapper mis-dispatching deprecated `str`-first filters (inspect_ai classifies on the *outermost* filter's first-param annotation, `_bridge/util.py:115`). Good fix — but this is now the only agent in the package whose `filter` rejects the deprecated form. If intentional, a docstring note would help.
- **`RunnerPayload` is defined three times** — a TypedDict in `antigravity.py`, a TypedDict in `sdk_runner.py`, and a pydantic model in `sdk_runner.py` — with `load_payload` hand-copying all seven fields. Because `google.antigravity` is imported at `sdk_runner.py` module scope, nothing in that file is host-testable, so no test can catch drift. Moving the SDK import inside `build_config`/`run` would make `load_payload` and payload-shape parity testable on the host.
- **The runner-hardening chown is not effective.** `runner.py` is chowned `root:root 0644` inside `/home/model/.antigravity`, which is `0700 model:model` — unlink permission comes from the directory, so the `model` user can replace the runner regardless of file ownership (and on the egress path the interpreter itself lives in the model-writable `/var/tmp` venv). Not a live exploit, but it reads as protection it doesn't provide: either move the runner to a root-owned directory or remove the hardening.
- Nits: the two dummy-key strings (`"inspect-bridge-unused"` in the env, `"inspect-bridge"` in the endpoint) could be one constant carried in the payload — as written, the credential-boundary test only covers one of them; and `bridge_base_url` could use `bridge.port` rather than the locally computed `bridge_port`, matching `gemini_cli`.

## Non-blocking: tests

- No coverage of `ensure_antigravity_sdk` / `_python_with_sdk` / `_base_python`, though `test_codex_agentbinary.py` and `test_ensure_wheels_install.py` establish the mock-sandbox pattern for exactly this. The blocker-4 probe fix should land with a regression test here.

## Non-blocking: docs (not addressed)

If this remains a public registered agent, the usual touch points:

- `docs/antigravity.qmd` (every other agent has a page)
- nav entry in `docs/_quarto.yml`
- agent-table row in `docs/index.qmd` (lines 43–48)
- `### antigravity` under `## Agents` in `docs/reference/index.qmd`
- optionally the `run_example` agent `Literal` in `tests/conftest.py` and the `examples/*/task.py` dispatchers

If the outcome of blocker 2 is "scoped adapter," this list shrinks but the scoping needs documenting somewhere regardless.

## Security

The central claim holds: only a dummy key enters the sandbox, real credentials stay host-side, loopback is excluded via `NO_PROXY`/`no_proxy`, and `PYTHONNOUSERSITE=1` blocks user-site injection in the runner. No injection risk in the `install -d` / `chown` f-strings, since `user` is validated equal to `model` before interpolation. The `root` execs match existing precedent in `_util/node.py` and `_util/ripgrep.py`, but will fail on sandboxes that block root exec — worth documenting. The provisioning path does an unpinned-hash `pip install` requiring egress, already flagged as follow-up.

sjawhar:
Claude here, on Sami's behalf. All four blockers addressed in the latest push, taking your "generalize" path on blocker 2 (Sami's explicit call):

1. **inspect_ai floor**: satisfied via the merge with main — `pyproject.toml` now declares `inspect_ai>=0.3.251`, above the 0.3.250 you asked for. The docstring on the runner's config builder also states the `functionResponse` re-roling dependency explicitly.
2. **Generalized**: the `taiga-mcp` literal is gone everywhere. New `mcp_servers` parameter; all configured servers (static + bridged) flow through the payload as a list; the policy predicate allows any configured server name; non-HTTP configs raise with a clear message (localharness reaches servers over streamable HTTP). `user`/`cwd` now default to the sandbox conventions via `resolve_agent_cwd`, matching gemini_cli; the `user`-interpolating shell strings are gone (argv-form execs, no root execs at all — see the hardening note below). Default-argument construction now runs (toolless) instead of raising.
3. **Conversation resume**: the runner reports the SDK `conversation_id` in its result line; the host persists it in the sample store and threads it into `LocalAgentConfig.conversation_id` on re-invocation. If a transcript already has assistant turns but no saved conversation exists, the agent raises loudly rather than silently truncating. Parsing is covered by host-side tests; if the runner output carries no id, a warning says resume won't be possible.
4. **Probe isolation**: `_python_with_sdk` and the post-install verify now run with `PYTHONNOUSERSITE=1`, with a regression test asserting the env on every probe exec.

Also taken from the non-blocking list: endpoint model carried through the payload as an `endpoint_model` parameter (fixes the inert `model_aliases` and removes the hidden constant); one dummy-key constant threaded through the payload for both env and endpoint; `bridge.port` used directly; `trace` gated behind a new `debug` parameter; `version` exposed (default 0.1.7); the SDK import moved inside `build_config()`/`run()` so `load_payload` is host-testable (with tests); the ineffective runner chown hardening removed with a comment explaining why (ownership can't protect a file in a model-writable directory, and the process runs as the model user anyway); the filter's Model-first typing documented; drift detection added — a declared tool outside allowlist ∪ the four known engine tools is still dropped (fail closed) but now logs a warning naming the drifted tools once; and the docs touch points (antigravity.qmd, nav, index table, reference) added.

Tool-confinement comment now also states explicitly that the filter supplements capability delegation (`enabled_tools=[]`) rather than replacing it, per your note.

`make check` clean (ruff, mypy strict) and the full suite passes (153). CI runs will need a workflow approval as with the other fork PRs.

src/inspect_swe/_antigravity/antigravity.py
```diff
@@ -0,0 +1,411 @@
...
+    server (static or bridged) must be an HTTP config with a URL.
+    """
+    entries: list[McpServerEntry] = []
+    for config in configs:
+        if not isinstance(config, MCPServerConfigHTTP) or not config.url:
+            raise ValueError(
+                f"antigravity requires HTTP MCP server configs with a URL; "
+                f"server {config.name!r} is {type(config).__name__}."
+            )
+        entries.append({"name": config.name, "url": config.url})
```
craig-meridian: `_mcp_server_entries` reduces every `MCPServerConfigHTTP` to `name` and `url`, discarding `tools`, `headers`, and `type`. `tools` is Inspect's explicit per-server allowlist; the runner then creates `McpStreamableHttpServer` without `enabled_tools` and adds a server-wide allow policy. Consequently, `tools=["read_only"]` silently exposes every tool on that server. Please carry the allowlist through the payload, map a list-valued `tools` to `enabled_tools`, and scope the policy with `policy.allow(server, mcp_tools=...)`; otherwise the wildcard policy still authorizes tools outside Inspect's allowlist. Preserve `headers` for authenticated servers. Also, `MCPServerConfigHTTP` represents both `type="http"` and `type="sse"`; an SSE config currently passes this check and is silently reconstructed as streamable HTTP, so either preserve that transport if supported or reject SSE explicitly. This is capability widening plus transport/auth breakage, and violates the confinement contract.

tests/test_antigravity.py
```diff
@@ -0,0 +1,133 @@
...
+    exec_mock = AsyncMock(
+        return_value=SimpleNamespace(success=success, stdout="", stderr="")
+    )
+    return SimpleNamespace(exec=exec_mock), exec_mock
+
+
+def test_sdk_presence_probe_runs_under_runner_import_isolation() -> None:
+    """A user-site-only install must not pass the probe (PYTHONNOUSERSITE=1)."""
+    sandbox, exec_mock = _sdk_probe_sandbox(success=True)
+    python = anyio.run(agentbinary._python_with_sdk, sandbox, "model")
```
craig-meridian: This test passes a `SimpleNamespace` to `_python_with_sdk`, whose parameter is annotated as `SandboxEnvironment`. Under the repository's strict mypy configuration this produces: `Argument 1 to "run" has incompatible type ...`. Please use a typed fake/cast or an async wrapper so the required type-check job passes.

src/inspect_swe/_antigravity/agentbinary.py
```diff
@@ -0,0 +1,132 @@
...
+    if python is not None:
+        return python
+    return await _provision_antigravity_sdk(sandbox, user, version)
+
+
+async def _python_with_sdk(
+    sandbox: SandboxEnvironment,
+    user: str | None,
+) -> str | None:
+    for python in _CANDIDATE_PYTHONS:
```
craig-meridian, line 74: The provisioned interpreter can never satisfy a later presence probe: `_python_with_sdk` checks `/opt/venv/bin/python`, `python3`, and `python`, but not `/var/tmp/.antigravity-sdk-venv/bin/python`. Consequently, every subsequent invocation on a generic image—including follow-up turns—re-enters `_provision_antigravity_sdk`, recreates the venv, and runs `pip install` again. Please probe and reuse the dedicated venv before provisioning, with a regression test showing that a second `ensure_antigravity_sdk` call does not reinstall. Since `version` is configurable, the reuse check should also avoid silently returning a different installed version.

src/inspect_swe/_antigravity/antigravity.py
```diff
@@ -0,0 +1,411 @@
...
+            bridged_tools=bridged_tools,
+        ) as bridge:
+            sbox = sandbox_env(sandbox)
+
+            # ensure google-antigravity is present (skip if baked into the image)
+            python = await ensure_antigravity_sdk(sbox, user, version=version)
+
+            # resolve working directory (home dir if sandbox default is '/')
+            agent_cwd = await resolve_agent_cwd(sbox, user, cwd)
+            data_dir = join_path(agent_cwd, _DATA_DIRECTORY_NAME)
```
craig-meridian: `resolve_agent_cwd()` returns the working directory, not necessarily the user's home; with an explicit `cwd`, it will normally be the repository checkout. Using it for `data_dir` and `HOME` writes `.antigravity/runner.py`, `request.json`, and session state into the evaluated repository and presents that repository as the user's home. This misses the `resolve_agent_cwd` plus separate `echo $HOME` split requested in the first review. Please keep `cwd=agent_cwd`, but resolve the actual sandbox home separately and place `HOME`, `app_data_dir`, `save_dir`, runner, and request state there—or in another dedicated non-workspace state directory.

craig-meridian (changes requested):
Thank you Sami, this is looking good. I've done another AI-assisted pass which has surfaced 4 comments. Let me know if you disagree with any. 

One of them could have been spotted at an earlier review round - apologies for missing this.

src/inspect_swe/_antigravity/antigravity.py
sjawhar: Fixed in 78b6157. `McpServerEntry` now carries `headers` and `tools` (None meaning "all"). The runner maps a list-valued `tools` to `enabled_tools` on `McpStreamableHttpServer`, scopes the per-server policy with `policy.allow(server, mcp_tools=...)` instead of the server-wide wildcard, and the `call_mcp_tool` dispatcher predicate now also rejects calls naming a tool outside the allowlist, so the allowlist holds at all three layers. `headers` are forwarded for authenticated servers.

SSE configs are rejected explicitly with a clear message rather than preserved: the pinned SDK has no SSE client (`McpStreamableHttpServer` is its only HTTP server type), so there is no transport to preserve.

Covered by new unit tests, and I verified construction against the real 0.1.7 SDK: a `tools=["read_only"]` server yields the policy chain `deny_all` + scoped dispatcher predicate + `secrets/read_only` (no `secrets/*`), while a `tools="all"` server keeps its wildcard.

tests/test_antigravity.py
sjawhar: Fixed in 78b6157. Reproduced the exact error under `mypy --strict` first. The probe fake is now `cast(SandboxEnvironment, SimpleNamespace(exec=...))`, and the spec kwargs helper moved to a TypedDict, which also let the two pre-existing `type: ignore[arg-type]`s in this file come out. mypy reports 0 issues across the package.

src/inspect_swe/_antigravity/agentbinary.py
sjawhar: Fixed in 78b6157. `_python_with_sdk` now probes the dedicated venv after the baked candidates, and that probe is version-checked (`importlib.metadata.version("google-antigravity") == version`) so a configurable `version` is never silently satisfied by a stale install; a mismatch falls through to provisioning, which pip-installs the requested pin into the same venv. The post-install verify runs the same version check. Regression tests cover both directions: a second `ensure_antigravity_sdk` call returns the venv python with no second install, and a version bump does reinstall.

src/inspect_swe/_antigravity/antigravity.py
sjawhar: Fixed in 78b6157. The agent now detects the sandbox home separately (`sh -c 'echo $HOME'`, the same detection gemini_cli uses) and places `HOME`, the data directory, runner, request, and `save_dir` there; `cwd` stays `agent_cwd`. `sdk_execution_spec` takes `home` as its own parameter, with a test asserting the exec env's HOME points at the sandbox home and not the workspace.

sjawhar:
Claude here, on Sami's behalf. All four comments addressed in 78b6157 (details in the threads):

1. **MCP allowlist/headers/transport**: `tools` and `headers` now ride the payload. The runner enforces the allowlist at three layers, all fail closed: `enabled_tools` on the server config, per-tool `policy.allow(server, mcp_tools=...)` instead of the server-wide wildcard, and a dispatcher predicate that rejects `call_mcp_tool` calls naming a tool outside the allowlist. SSE configs are rejected explicitly, since the pinned SDK has no SSE client to hand them to.
2. **Test typing**: typed `cast` for the probe fake plus a TypedDict for the spec kwargs (which also removed this file's two pre-existing `type: ignore`s).
3. **Venv reuse**: the presence probe now includes the dedicated venv, version-checked so a stale install never silently satisfies a different requested `version`. Regression tests assert no reinstall on a second call and a reinstall on a version bump.
4. **Home/cwd split**: sandbox home detected separately (gemini_cli's `echo $HOME` pattern); SDK state and `HOME` live there, `cwd` stays the resolved working directory.

Verification: `ruff check` and `ruff format --check` clean, `mypy --strict` 0 issues (110 files), antigravity tests 19/19 (was 7, now includes the new allowlist/headers/SSE/venv-reuse/home tests). I also constructed the confined config against the real `google-antigravity==0.1.7` to confirm the policy chain and the `headers`/`enabled_tools` pass-through, and exercised the dispatcher predicate directly (allowlisted tool approved, non-allowlisted tool on the same server rejected, unknown server rejected). Full suite: 158 passed; the one failure is `tests/test_k8s.py`, which fails in my environment on inspect_ai's anthropic-package minimum-version gate before the test body runs, unrelated to this change.

sjawhar:
Claude here, on Sami's behalf. One more commit on this branch (`f2dfc7e`), and it is a functional fix rather than a review response, so calling it out explicitly.

## view_file is now allow-listed (confined to the agent's data dir)

localharness offloads any large MCP tool result to a file and returns the model only a pointer ("The output was large and was saved to: ..."). With `enabled_tools=[]` the model had no builtin that could read that file back, so **every tool observation above localharness's internal cap (~4KB) was silently lost** — no error, the model just never saw the content. On browser-surface tasks that is most of the page text.

The fix re-enables exactly one builtin, `VIEW_FILE`, across the three layers that stripped it (the capabilities allowlist, the localharness policy, and the GenerateFilter tool-name allowlist). It is confined by a policy predicate to the agent's own `app_data_dir`, where localharness writes the offloaded results: it cannot read task files, the workspace, or grading ground truth. Paths are lexically normalized first so `..` cannot escape, and the prefix match is segment-aware so a sibling directory does not match. The predicate is a plain host-testable factory (`_confine_to_dir`) with unit tests for traversal, sibling-prefix, non-string and missing-argument cases.

## End-to-end evidence

Ran the agent against a real sandbox (Gemini 3.6 Flash, local Docker, SDK baked at the pinned 0.1.7):

| run | result | view_file reads | offloaded payloads | model calls |
|---|---|---|---|---|
| tool-calling smoke | regular 1.00 | n/a | 0 | 14 |
| browser task, **with** fix | regular 1.00 | 6 | 3 recovered | 32 |
| browser task, **without** fix (control) | regular 1.00 | 0 | **4 lost** | 96 |
| injection task, 3 epochs, with fix | **injection 1.00 (3/3)** | 9–12 per sample | 6–9 per sample | 334 |

The control is the interesting row: with view_file removed the agent still completed that particular benign task, but blind — 4 offloaded payloads never read, 5x the MCP calls, 3.3x the tokens, 3x the wall time, repeatedly clicking without page context (`Element reference, coordinates, or selector are required`). On an injection task the offloaded page text is precisely where the injected content lives, so losing it suppresses the behavior under test rather than merely slowing the agent down.

Confinement held in every run: 0 view_file paths outside the data directory across all samples.

`make check` clean (ruff + mypy strict, 0 issues / 110 files); antigravity tests 20/20.

src/inspect_swe/_antigravity/sdk_runner.py
```diff
@@ -0,0 +1,235 @@
...
+            enabled_tools=[types.BuiltinTools.VIEW_FILE],
+            enable_subagents=False,
+        ),
+        mcp_servers=mcp_servers,
+        policies=[
+            policy.deny_all(),
+            policy.allow("call_mcp_tool", when=_targets_configured_server),
+            policy.allow(
+                types.BuiltinTools.VIEW_FILE.value,
+                when=_confine_to_dir(payload["app_data_dir"]),
```
craig-meridian: [P1] Keep MCP credentials outside view_file's readable tree

`view_file` is authorized for every path under `app_data_dir`, but the host writes `request.json` into that same directory, and the payload's `mcp_servers` entries include `headers`. For an authenticated MCP server, the model can therefore call `view_file` on `$HOME/.antigravity/request.json` and read or exfiltrate its `Authorization` header. This violates the stated invariant that `view_file` can only read offloaded tool results. Confirmed against `google-antigravity==0.1.7`: running this config's own `policy.enforce` hook over `view_file{"AbsolutePath": "$HOME/.antigravity/request.json"}` returns `allow=True`.

Please either confine `view_file` to a dedicated offload-results subtree, or move the runner/config—and any session files that may serialize the configuration—outside its readable tree. Please add a regression test proving that `request.json` and an auth-bearing configuration cannot be read.

Note `save_dir` is `app_data_dir/session`, so relocating only `request.json` may still leave configuration-bearing session files readable. The Python SDK sends MCP headers to the compiled harness over its local websocket rather than intentionally writing them itself, but what the harness persists under `storage_directory` is not inspectable. Suggested layout: `runner.py`, `request.json`, and `save_dir` under a sibling `$HOME/.antigravity-state`, leaving `app_data_dir` holding only harness-written data. That avoids having to know which subdirectory the offloads land in.

src/inspect_swe/_antigravity/sdk_runner.py
```diff
@@ -0,0 +1,235 @@
...
+    match. Host-testable: no SDK import.
+    """
+    confined = posixpath.normpath(directory)
+
+    def _within(args: dict[str, Any]) -> bool:
+        raw = args.get("AbsolutePath")
+        if not isinstance(raw, str) or not raw:
+            return False
+        resolved = posixpath.normpath(raw)
+        return resolved == confined or resolved.startswith(confined + "/")
```
craig-meridian: [P2] Resolve symlinks before enforcing the view_file boundary

`_confine_to_dir` only applies `posixpath.normpath`, so a path that is lexically beneath `app_data_dir` can still resolve outside it through a symlink. For example, if `.antigravity/offload` points at `/workspace/repo/grading`, then `.antigravity/offload/secret.txt` passes this predicate while `view_file` opens the file outside the allowed directory. The SDK runs as the model user and this directory is model-writable, so this bypasses the claimed confinement boundary.

Please compare resolved filesystem paths or reject symlink components, and add a regression test covering a symlink inside the allowed directory that targets a file outside it.

The SDK already ships this pattern: `hooks/policy._secure_normalize_path` (`Path.resolve()`, `OSError` → fail closed) plus `_is_path_in_workspace` for segment-aware containment—worth mirroring rather than reinventing. Two caveats: resolving both sides hardens the boundary but does not close it, since the predicate runs in the SDK process while the open happens in localharness, so the docs should say `view_file` is confined to `app_data_dir` rather than that it cannot read elsewhere. The regression test needs real files, so use `tmp_path` rather than literal `/home/model/...` paths. Exploiting this in practice requires an MCP tool that can plant a symlink but cannot read the target directly; with a shell in the environment, the symlink route adds nothing. This is therefore hardening, but the fix is small enough to include.

src/inspect_swe/_antigravity/sdk_runner.py
```diff
@@ -0,0 +1,235 @@
...
+    view_file exists ONLY to read back localharness's offloaded tool results,
+    which live under the agent's own app_data_dir. Paths are lexically
+    normalized first so ``..`` traversal cannot escape the prefix, and the
+    prefix match is segment-aware so a sibling like ``<dir>-evil`` does not
+    match. Host-testable: no SDK import.
+    """
+    confined = posixpath.normpath(directory)
+
+    def _within(args: dict[str, Any]) -> bool:
+        raw = args.get("AbsolutePath")
```
craig-meridian: [P2] Surface path-argument drift (non-blocking—follow-up is fine)

This keys off `args["AbsolutePath"]` with no drift signal. If the harness renames or URI-wraps that key, `_within` returns `False` for every call and `view_file` silently becomes deny-all—offloaded results become unreadable again, this time through denial loops.

The SDK cannot cover for it either. `policy.workspace_only()` matches on `ToolCall.canonical_path`, which `hook_router` derives only from `{path, file_path, directory_path, TargetFile}`. Consequently, `canonical_path` is `None` for `view_file`, making the built-in path confinement a no-op for it. Hand-rolling the predicate is right; it just needs the same one-shot drift warning that `_confine_declared_tools` provides for unknown tool names.

The warning must reach a host-visible channel: `sdk_runner` stderr is only surfaced when `debug=True` or when the runner fails, so a normal `logger.warning` inside the sandbox runner would remain invisible during successful default runs.

docs/antigravity.qmd
```diff
@@ -0,0 +1,44 @@
...
+Runs Google's Antigravity SDK (`google-antigravity`, which bundles the `localharness` engine) headless inside an Inspect sandbox. Model calls are routed through the sandbox agent bridge: the SDK's native connection speaks the Gemini `generateContent` wire directly to the bridge via a `GeminiAPIEndpoint` `base_url` override (no OpenAI translation), so only a placeholder API key ever enters the sandbox.
+
+## Options
+
+The following options are supported for customizing the behavior of the agent:
+
+| Option | Description |
+|------------------------------------|------------------------------------|
+| `system_prompt` | Additional system prompt to append to default system prompt. |
+| `mcp_servers` | MCP servers to make available to the agent. localharness reaches servers over streamable HTTP, so `type="http"` configs are required and `type="sse"` is rejected rather than silently reconnected. |
```
craig-meridian: [P3] "silently reconnected" → "reconstructed".

craig-meridian (changes requested):
Thanks Sami — the earlier review items are addressed. The newly added view_file path still has one blocking credential-boundary issue; the remaining inline comments are hardening or non-blocking follow-ups.

src/inspect_swe/_antigravity/sdk_runner.py
sjawhar: Fixed in 58959f2d. Reproduced it first: with the old layout, running the config's own policy hook over `view_file{"AbsolutePath": "<app_data_dir>/request.json"}` returns allow=True, exactly as you found.

Took your suggested layout. `app_data_dir` now holds only harness-written data; the runner, `request.json` and `save_dir` move to a sibling `$HOME/.antigravity-state`. That avoids needing to know which subdirectory the offloads land in, and it covers the session-file concern you raised — since what the compiled harness persists under `storage_directory` isn't inspectable, keeping `save_dir` out of the readable tree is the only assumption-free answer.

Verified against `google-antigravity==0.1.7` by building this agent's real config and calling the resulting `view_file` policy's own predicate:

| path | allow |
|---|---|
| offloaded result in `app_data_dir` | True |
| `request.json` (Authorization header) | False |
| `runner.py` | False |
| session store under `save_dir` | False |
| symlink in `app_data_dir` → state dir `request.json` | False |

Regression tests cover all of it, including an auth-bearing `request.json` and the session path (`test_view_file_predicate_cannot_read_the_request_or_session_state`), using `tmp_path` with real files rather than literal `/home/model` paths.

Worth noting for scope: bridged servers get no `headers` from inspect_ai's bridge, so the live exposure needed a caller-supplied authenticated static server. The boundary was still wrong, and `request.json` also carries the prompt and system instructions.

src/inspect_swe/_antigravity/sdk_runner.py
sjawhar: Fixed in 58959f2d, mirroring the SDK helper you pointed at rather than reinventing it: `_resolve_strictly` is `Path(path).resolve()` with `OSError` denying the call (the same non-strict resolve + fail-closed contract as `hooks/policy._secure_normalize_path`), applied to both the requested path and the confined directory, and containment compares resolved path parts so a sibling like `<dir>-evil` still cannot match. If the confined directory itself cannot be resolved, every call is denied.

Regression test `test_view_file_predicate_rejects_symlink_escape` plants a real symlink inside the allowed directory pointing at an outside directory holding a `grading/regular.py`, and asserts both the target file and the link itself are denied — with `tmp_path`, as you noted the test needs real files.

Also took your docs point: the wording now says `view_file` is confined to `app_data_dir` rather than claiming it cannot read elsewhere, and states explicitly that the predicate runs in the SDK process while the open happens in localharness.

src/inspect_swe/_antigravity/sdk_runner.py
sjawhar: Addressed in 58959f2d rather than deferred — the failure mode you describe (view_file becomes deny-all, offloaded output unreadable again, this time through denial) is the same bug the read-back path exists to fix, so it seemed worth closing now.

The key is a named constant `_VIEW_FILE_PATH_ARG`, and when a call arrives without it the predicate records that call's argument keys. Your point about the channel is the reason it is not a `logger.warning`: the runner's stderr only reaches the host with `debug=True` or on failure, which excludes exactly the successful-but-degraded run. The drifted keys instead ride the runner's JSON result line (`view_file_unexpected_arg_keys`), which the host already parses for the conversation id, and the host logs a warning naming them and saying large tool output is not reaching the model.

Tests: `test_view_file_predicate_records_path_argument_drift` for the recording, `test_reported_view_file_arg_drift_reads_the_result_line` for the host-side parse.

Thanks for the `canonical_path` detail — I hadn't checked `hook_router`, and it confirms `policy.workspace_only()` is a no-op for `view_file`, so the hand-rolled predicate has to carry the boundary itself.

docs/antigravity.qmd
sjawhar: Fixed in 58959f2d — "reconstructed".

sjawhar:
Claude here, on Sami's behalf. All four addressed in `58959f2d` (details in the threads).

**P1, the blocking one.** Reproduced your finding first — with the old layout the config's own policy hook returns allow=True for `view_file` on `<app_data_dir>/request.json`. Took your suggested layout: `app_data_dir` now holds only harness-written data, and the runner, `request.json` and `save_dir` move to a sibling `$HOME/.antigravity-state`. Re-verified against `google-antigravity==0.1.7` through the real policy predicate:

| path | allow |
|---|---|
| offloaded result in `app_data_dir` | True |
| `request.json` (Authorization header) | False |
| `runner.py` | False |
| session store under `save_dir` | False |
| symlink in `app_data_dir` → state dir `request.json` | False |

**P2 symlinks.** Mirrored `hooks/policy._secure_normalize_path` — non-strict `Path.resolve()` on both sides, `OSError` fails closed, containment on resolved parts. Test plants a real symlink to an outside `grading/regular.py` and asserts denial, using `tmp_path`.

**P2 drift.** Closed rather than deferred, since a renamed path argument reintroduces the exact bug `view_file` was added to fix. Drifted argument keys ride the runner's JSON result line (the host already parses it) and the host logs the warning — not in-sandbox, for the channel reason you gave.

**P3.** "reconstructed", and the docs now say `view_file` is *confined to* `app_data_dir` rather than claiming it cannot read elsewhere, noting the predicate runs in the SDK process while the open happens in the harness.

`make check` clean (ruff + mypy strict, 110 files); full suite 163 passed / 63 skipped.

Two disclosures. First, this branch also picked up a docs commit (`b21e612a`) before your review landed, correcting a stale `enabled_tools=[]` claim in `antigravity.qmd`. Second, on the scope of P1 as shipped: bridged servers get no `headers` from inspect_ai's bridge, so reading a real credential required a caller-supplied authenticated static server — the boundary was still wrong, and `request.json` also carries the prompt and system instructions.

craig-meridian (approved):
Thanks very much Sami, approved!

PR #95, opened 2026-07-24
fix(codex_cli): refresh bundled model catalog snapshot to track live latest

sjawhar:
Hi, I am Claude acting on Sami Jawhar (sjawhar)'s behalf.

This refreshes the bundled Codex fallback catalog from `openai/codex` `rust-v0.145.0` (July 2026), including the current `gpt-5.6-sol` latest entry and its live priorities. It keeps the fallback aligned when the version-matched catalog cannot be fetched.

Validation:
- `uv run pytest tests/test_codex_agentbinary.py::test_bundled_catalog_tracks_live_latest -v`
- `uv run pytest tests/test_codex_agentbinary.py tests/test_codex_model_catalog.py tests/test_codex_align.py -q`
- Ruff and mypy on the changed snapshot module

Separate observation: the drift test docstring says it is github-action-gated, but the test body only skips when its live fetch fails. A newly shipped model can therefore hard-fail CI until the snapshot is refreshed. I can send a separate follow-up if maintainers would like that gate implemented; this PR intentionally only refreshes the snapshot.

### DIFF
--- a/src/inspect_swe/_codex_cli/_bundled_catalog.py
+++ b/src/inspect_swe/_codex_cli/_bundled_catalog.py
@@ -8,9 +8,9 @@ trimmed snapshot of the fields the resolver reads
 (``slug``/``priority``/``apply_patch_tool_type``/``supports_search_tool``) is
 sufficient.
 
-Snapshot source: ``openai/codex`` ``codex-rs/models-manager/models.json`` (main,
-June 2026). Refresh when bumping the default Codex version; the live fetch keeps
-this exact when ``raw.githubusercontent.com`` is reachable.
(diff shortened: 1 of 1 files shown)


craig-meridian (approved):
Thanks a lot for this, looks great.

On the "separate observation" in the description, I may be misreading your point, but `test_bundled_catalog_tracks_live_latest` looks like it's already decorated with `@skip_if_github_action`, which skips it whenever `GITHUB_ACTIONS` is set. So I'd have thought a newly shipped model wouldn't cause CI (in the GitHub action sense). Though it could of course fail pytest locally.

PR #98, opened 2026-07-27
feat: support replacing Claude Code system prompt

jannes-apollo:
Drafted by GPT-5

## Summary

- preserve `system_prompt`'s existing append behavior
- add `replace_system_prompt` for Claude Code's `--system-prompt` replacement behavior
- raise a clear runtime error when both options are specified
- document and test the append, replace, and resume behavior

## Testing

- `uv run --extra dev pytest tests/test_claude_code_system_prompt.py tests/test_claude_code_mcp_allowed_tools.py -q` (15 passed)
- `uv run --extra dev -- make check`


epatey:
@review

docs/claude_code.qmd
```diff
@@ -15,7 +15,8 @@ The following options are supported for customizing the behavior of the agent:
 
 | Option | Description |
 |------------------------------------|------------------------------------|
-| `system_prompt` | Additional system prompt to append to default system prompt. |
+| `system_prompt` | System prompt that replaces Claude Code's built-in prompt. |
+| `append_system_prompt` | Additional instructions to append to the effective system prompt. |
```
epatey: If I'm reading this correctly, this is a breaking change. Aren't you changing the meaning of `system_prompt`? In production, it currently appends. With this PR, it will replace.

epatey:
Please add a `CHANGELOG.md` entry.

docs/claude_code.qmd
jannes-apollo: Yes it's a breaking change. In the claude code CLI, --system-prompt replaces, and --append-system-prompt appends. I do think this breaking change also aligns the implementation with intuition

jannes-apollo:
> Please add a `CHANGELOG.md` entry.

The changelog.md seems to be auto generated now. Or do you want me to highlight that this is a breaking change?

docs/claude_code.qmd
epatey: My thinking is that if people who pass `system_prompt` get this new code their evals will behave quite differently. If they're not release note readers, it may be quite confusing for them to debug. We want to avoid surprising behavior like that. So, I think we have to keep `system_prompt` the way it was and add `replace_system_prompt` - even though that doesn't strictly match claude itself.

Stepping back, I also think it would provide better DX if we used overload signatures to make it the nonsensical combination of passing both the append and replace options impossible.

epatey:
> > Please add a `CHANGELOG.md` entry.
> 
> The changelog.md seems to be auto generated now. Or do you want me to highlight that this is a breaking change?

whoops. I forgot. good point

docs/claude_code.qmd
jannes-apollo: > Stepping back, I also think it would provide better DX if we used overload signatures to make it the nonsensical combination of passing both the append and replace options impossible.

I added a check for this. The overloading signatures option seemed a bit overkill to me, but I don't have strong opinions here

epatey (approved):
Thank you

PR #100, opened 2026-07-28
feat(bridge): add transparent_proxy to codex_cli and claude_code

sjawhar:
## Summary

Adds an opt-in `transparent_proxy` mode to the Codex CLI and Claude Code bridges. When enabled, the bridge forwards the client's own generation parameters (via inspect_ai's `forward_generation_config`) instead of clearing them, so an evaluated agent's internal model calls reproduce the configuration the agent's SDK actually sent. The bridge alias table and fallback model are retained; this does not replace them.

## Notes

- `transparent_proxy` composes with `auto_review`. The Codex guardian slug (`codex-auto-review`) is bound via `default=get_model(model)` so it resolves under the flag instead of losing its alias.
- Generation params are forwarded bridge-wide, but the eval's explicitly-set `GenerateConfig` still takes precedence on the active model (forwarding only fills fields the eval leaves unset). In practice this changes the internal reviewer/classifier/guardian calls and otherwise-unset params.
- Opt-in: `transparent_proxy` defaults to `False`.

## Verification

- `KUBECONFIG=/dev/null uv run pytest tests/ -x -q` (140 passed, 63 skipped)
- `uv run ruff format --check .`
- `uv run ruff check .`
- `uv run mypy src tests`

Written by Claude, acting for Sami (@sjawhar).

### DIFF
--- a/src/inspect_swe/_claude_code/claude_code.py
+++ b/src/inspect_swe/_claude_code/claude_code.py
@@ -129,6 +129,7 @@ def claude_code(
     model: str | None = None,
     model_config: str | None = None,
     model_aliases: dict[str, str | Model] | None = None,
+    transparent_proxy: bool = False,
     opus_model: str | None = None,
     sonnet_model: str | None = None,
     haiku_model: str | None = None,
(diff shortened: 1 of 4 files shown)


epatey (changes requested):
Thanks for the PR! I can't get `transparent_proxy=True` to route anything, so I want to understand how you ran it.

With the flag on there's no fallback and no alias table, so every request resolves as `get_model(<raw name the agent sent>)` — and both agents send bare names (codex its `--model` slug, claude_code the presented name), which `get_model()` rejects:

```python
>>> resolve_inspect_model("gpt-5.1-codex", None, None)
ValueError: Model name 'gpt-5.1-codex' should be in the format of <api_name>/<model_name>.
```

Same for `codex-auto-review` and `claude-sonnet-4-5-…` (inspect_ai 0.3.250). In particular `auto_review=True, transparent_proxy=True` can't work: the guardian slug is hardwired in the codex binary and this PR drops the alias that routed it. The new test only constructs the agent, so it doesn't exercise any of this.

Did you run this with a provider-prefixed `model_config`, task model roles named after the slugs, or something else? That'll tell us whether the fix is docs + a guard, keeping the guardian alias under the flag, or provider inference upstream in inspect_ai.

sjawhar:
You're right, and I could reproduce it locally. With transparent_proxy=True I dropped model_aliases and the fallback model entirely, so every request resolved through get_model(<raw name the client sent>). Both codex_cli (its own --model slug) and claude_code (the presented identity) send bare names, and codex-auto-review is a slug Codex hardcodes into every guardian request regardless of configuration, so none of it had a chance to resolve. The construction-only test never executes the bridge, so it never caught this.

Fixed by keeping the alias table and fallback model in place for both agents. For the guardian, I went with your second option: resolve_codex_auto_review_model_aliases now gets default=get_model(model) when transparent_proxy=True, so codex-auto-review binds to the real target model even when CodexAutoReview() doesn't set its own model. This mirrors the pattern the ACP codex bridge already uses for its no-fallback setup. Codex's ordinary --model slug keeps resolving through the fallback exactly as it does when transparent_proxy=False.

For claude_code, the presented identity and every opus/sonnet/haiku/subagent role were already aliased to their served Model regardless of transparent_proxy, so dropping the alias table there wasn't fixing anything, only breaking resolution. transparent_proxy now only changes forward_generation_config, which is the actual fix for the auto-mode classifier's dropped generation parameters.

I replaced the construction-only test with ones that exercise the real resolution path: resolve_codex_auto_review_model_aliases and resolve_claude_code_models feeding into resolve_inspect_model, covering the guardian slug with no explicit model and the bare presented/slug names you flagged. Pushed.

epatey (approved):
This is clean now: the guardian binding, retained alias table and fallback, and the tests all check out. I brought the description up to date to match the final design. Approving.

PR #101, opened 2026-07-28
fix(mcp): wait for bridged MCP endpoints before agent launch, and fail loudly if unreachable

sjawhar:
## Problem

Agents that receive bridged MCP tools read their MCP config **at startup**, while the bridge proxy comes up **asynchronously**. Launch before the proxy is listening and the agent has no bridged tools — and **nothing errors**. It reports `No such tool available`, emits an empty response, and that response is scored as an ordinary toolless trajectory. No error field means `--retry-on-error` never fires and reuse logic treats the sample as valid indefinitely.

Verbatim from an affected transcript (`[6]` succeeding rules out provisioning — the tool existed, then the restart lost it):

```
[4]  assistant WaitForMcpServers        -> ready: true
[6]  assistant mcp__taiga-mcp__browser  -> "Successfully navigated to ..."   ← WORKS
[8]  assistant mcp__taiga-mcp__browser  -> "The operation timed out."
[10] user      "Continue from where you left off."
[12] user      <system prompt re-injected>                                   ← restart
[13] assistant mcp__taiga-mcp__browser  -> "No such tool available: mcp__taiga-mcp__browser"
[15] assistant (empty)
```

**Measured: 209 affected samples across six eval arms, a session restart preceding every one.** `claude_code` is most exposed because its retry loop relaunches with `--resume`, and three separate triggers reach it (tool timeout, unavailable auto-mode classifier, and the `uncaught_error_count` path) — matching the four inputs to `is_resume`.

## Why this is a gap, not a new feature

The readiness primitive already existed for exactly this reason. `acp/agent.py` polled the bridge endpoint before starting a session, commenting that agents connecting synchronously "**will silently skip tools if the proxy isn't ready yet**". No other agent gained the guard.

## Change

- `_util/mcp_ready.py` — shared `wait_for_mcp_endpoints`, which **raises `MCPEndpointsUnreachableError`** on timeout by default. `required=False` opts out.
- Gate every agent that consumes `bridge.mcp_server_configs` and launches its own subprocess: `_claude_code` (inside its retry loop, so `--resume` relaunches are covered), `_codex_cli`, `_kimi_code`, `_gemini_cli`, `_opencode`.
- `acp/agent.py` moves from proceed-anyway to `required=True`.
- `acp/_agents/*` are exempt **by design**: their MCP connection happens in the ACP `new_session`, which `acp/agent.py` already gates.

**Failing is the point.** Proceeding past an unreachable endpoint is what produced the bad samples. An errored sample is retryable and honest; a scored toolless one silently corrupts results.

## What I got wrong first, and one suspicion that didn't hold

The first commit here was incomplete, and the second fixes it — kept as two commits so the reasoning is visible:

- It **discarded the return value**, so on timeout it logged "proceeding anyway" and still launched toolless. It widened the race window but did nothing in the case that actually matters.
- It covered **only `claude_code`** — the one agent I had a transcript for — leaving four siblings with identical exposure.

A third suspicion **did not hold up** and is deliberately unchanged: the `curl -sf -X POST` probe is valid. The bridge proxy answers JSON-RPC over HTTP and returns `200` even for protocol-level errors, so a bodyless POST to a live endpoint yields `200` and `-sf` succeeds. No false negatives, so I left the probe as inherited rather than rewriting it as an `initialize` handshake.

## Tests

`tests/test_mcp_ready.py`: reachable, polls-then-ready, **raise-on-timeout**, `required=False` opt-out, empty-configs no-op, plus two structural guards —

1. the `claude_code` await sits between `consumer.reset()` and `exec_remote`, i.e. **inside** the retry loop;
2. **any** module using `bridge.mcp_server_configs` without waiting fails the suite, so a future agent can't reintroduce this.

Guards verified by reverting: each of the four agents individually fails guard 2, and removing the `claude_code` gate fails guard 1. Async is driven via `anyio.run` (already a runtime dep) — no new test dependency or pytest config.

`ruff check` clean · `ruff format` clean · `mypy src` clean (75 files) · `pytest tests` **121 passed, 62 skipped**.

## Not verified

The mechanism is established by code inspection plus 209 consistent transcripts, **not** by reproducing the race under instrumentation. I have not confirmed the endpoint is unreachable at the moment of relaunch. Two alternatives I could not exclude: `--resume` may restore MCP state from the persisted session rather than re-reading `--mcp-config` (in which case waiting cannot help that path), and the preceding tool timeout may have wedged the proxy rather than it being slow to start. In both cases this change is still correct — it converts a silently-scored toolless sample into a loud error — but it may not eliminate the underlying race. Flagging so a maintainer with a reproduction can judge.

### DIFF
--- a/src/inspect_swe/_claude_code/claude_code.py
+++ b/src/inspect_swe/_claude_code/claude_code.py
@@ -15,7 +15,13 @@ from inspect_ai.agent import (
 )
 from inspect_ai.model import ChatMessageSystem, GenerateFilter, Model, StopReason
 from inspect_ai.scorer import score
-from inspect_ai.tool import MCPServerConfig, Skill, install_skills, read_skills
+from inspect_ai.tool import (
+    MCPServerConfig,
+    MCPServerConfigHTTP,
+    Skill,
+    install_skills,
+    read_skills,
+)
 from inspect_ai.util import (
     ExecRemoteStreamingOptions,
     StoreModel,
@@ -39,6 +45,7 @@ from inspect_swe._claude_code._events.stream import (
     claude_code_event_stream,
 )
 from inspect_swe._util.centaur import CentaurOptions, run_centaur
+from inspect_swe._util.mcp_ready import wait_for_mcp_endpoints
 from inspect_swe._util.path import join_path
 
 from .._util._async import is_callable_coroutine
@@ -305,6 +312,13 @@ def claude_code(
             static_mcp_servers = list(mcp_servers or [])
             bridged_mcp_servers = bridge.mcp_server_configs
             all_mcp_servers = static_mcp_servers + bridged_mcp_servers
+            # HTTP endpoints we must confirm are live before EVERY launch: the
+            # bridge proxy starts asynchronously, and Claude Code reads
+            # --mcp-config at startup. If the proxy isn't listening yet the
+            # agent comes up with no MCP tools and reports NO error.
+            http_mcp_configs = [
+                c for c in all_mcp_servers if isinstance(c, MCPServerConfigHTTP)
+            ]
             if all_mcp_servers:
                 mcp_server_args, _ = resolve_mcp_servers(all_mcp_servers)
                 cmd.extend(mcp_server_args)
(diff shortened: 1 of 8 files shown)


sjawhar:
Converting to draft — self-review found this incomplete in three ways, so please don't merge as-is:

1. **It preserves the defect it claims to fix.** The call site discards the return value, so on timeout `wait_for_mcp_endpoints` logs "proceeding anyway" and the agent still launches toolless and still gets graded. This widens the race window (helps when the proxy is merely slow) but does nothing when it is actually unreachable — which is the case that produces the silent bad sample.

2. **It is too narrow.** Four other non-ACP agents consume `bridge.mcp_server_configs` and launch their own subprocess with the same exposure and no guard: `_codex_cli`, `_kimi_code`, `_gemini_cli`, `_opencode`. Fixing only `claude_code` patches the one case I had a transcript for.

3. **The readiness probe is unvalidated.** `curl -sf -X POST <url>` with no body: `-sf` fails on 4xx, so a live MCP server that rejects a bodyless POST would report unreachable. Inherited from the ACP path without testing.

Reworking to fail loudly on unreachable required tools, cover all affected agents from one choke point, and use a real MCP handshake as the probe. Will un-draft once that lands with evidence.

sjawhar:
Consolidated into https://github.com/meridianlabs-ai/inspect_swe/pull/106, which carries this PR's readiness gate as its base plus the Claude Code blocking-connect fix — one PR for the whole failure mode. (Posted by Claude on Sami's behalf.)

PR #106, opened 2026-07-31
fix: gate agent launch on bridged MCP readiness and block Claude Code's first model call on MCP connect

sjawhar:
## Scope

This PR retains the shared bridged-MCP readiness gate and the corrected curl image prerequisite documentation. The configurable readiness timeout and Codex slow-sandbox startup-timeout work have been split into a focused follow-up PR.

The readiness gate waits for bridged MCP endpoints to serve expected tools before launch, checks every bridged endpoint, fails loudly when the probe cannot run or readiness expires, and is applied to self-launching agents and ACP. Claude Code also retains its blocking MCP connection defaults.

## Verification

Claude ran the targeted regression suite covering readiness behavior, Claude Code environment defaults, and Codex MCP configuration: 59 passed. Claude also ran ruff on the retained changed paths: passed. The repository-wide mypy gate is currently blocked by nine existing `web_search` argument-type errors on upstream main, outside this PR.

## Split

The two requested commits are no longer in this branch. The follow-up PR contains the configurable `mcp_ready_timeout` and Codex startup-timeout work.


sjawhar:
Pushed two additions after investigating layer-2 coverage for the non-Claude agents (details now in the PR body's coverage table):

1. Codex: bridged MCP servers are now written with `required = true`, so codex (>= 0.119.0) blocks session init on their initialize+tools/list and `codex exec` fails loudly instead of running the turn without tools. Older codex versions ignore the key. Static caller-provided servers stay optional, mirroring the readiness-gate scoping.
2. Merged current main (conflicts were import/append unions with the auto_review work).

Gemini CLI and OpenCode turned out to already block the first turn on MCP connect in the headless modes we launch them in, so no changes there. Kimi CLI has no blocking knob at all (fire-and-forget connect in every mode, documented as by-design) — recorded as a known limitation in the body and a code comment.

This is Claude, working on Sami's behalf.

sjawhar:
Claude here, on Sami's behalf. Pushed one small commit to this branch (`00a13d40`), found while integrating this branch with #94 in our fork's release:

`fix(test): MCPServerConfigHTTP carries url, not the base MCPServerConfig` — the two tests added with the `required = true` work constructed `MCPServerConfig(type="http", name=..., url=...)`, but `url` is declared on `MCPServerConfigHTTP`, not the base class, so `mypy --strict` fails with `Unexpected keyword argument "url"`. The CI run on this PR only exercised `lint / lint` (the type-check job needs a workflow approval), so it went unnoticed. Switched both constructor sites and the import to `MCPServerConfigHTTP`; behavior is unchanged.

Verified on this branch: `ruff check` and `ruff format --check` clean, `mypy --strict` 0 issues (108 files), `pytest tests/test_codex_config.py` 30 passed.

sjawhar:
Claude here, on Sami's behalf. Rebased onto current main and added one commit (`eb1cc897`), because #94 merging changed what this PR owes.

`test_every_self_launching_bridged_agent_gates_on_mcp_readiness` asserts that every agent consuming `bridge.mcp_server_configs` awaits `wait_for_mcp_endpoints`. The antigravity agent does consume them and did not await the gate — it just wasn't in the tree when this branch was written. Now that #94 has landed, merging main in makes this branch's own test fail:

```
E  assert not ['_antigravity/antigravity.py']
tests/test_mcp_ready.py:365: AssertionError
```

So the gate is added to antigravity here, in the PR that introduces the contract and its enforcement, rather than anywhere else. Same shape as `gemini_cli`: bridged HTTP configs are awaited with `required=True`, static caller-provided servers stay ungated. Without it the SDK agent can launch before the bridge proxy serves a tool listing and run the turn with no environment tools and no error — a toolless trajectory scored as an ordinary one, which is the failure this PR exists to close.

`make check` clean (ruff + mypy strict, 114 files); `tests/test_mcp_ready.py` 14 passed; full suite 186 passed / 82 skipped.

src/inspect_swe/_util/mcp_ready.py
```diff
@@ -0,0 +1,190 @@
...
+    # Wall-clock deadline: each probe's own runtime (up to _PROBE_MAX_TIME while
+    # curl waits on a cold endpoint) counts against the budget. Accumulating the
+    # sleep interval instead would let a "120s" timeout stretch to roughly
+    # timeout/interval * _PROBE_MAX_TIME of real time when probes hang.
+    start = anyio.current_time()
+    deadline = start + timeout
+
+    while True:
+        remaining = deadline - anyio.current_time()
+        if remaining <= 0:
```
craig-meridian: [P2] Recompute the remaining deadline before each endpoint probe

`remaining` is calculated once before iterating over every pending endpoint. If several endpoints are slow, each receives a timeout based on the same stale value, so the loop can run well beyond the advertised wall-clock deadline. The `max(1, ...)` also permits a one-second probe when less than one second remains.

Please recompute `remaining` inside this loop and stop probing once it reaches zero (ideally also bounding `sbox.exec` with the exact remaining time). The regression test currently uses only one endpoint, so it does not catch this case.

src/inspect_swe/_kimi_code/kimi_code.py
```diff
@@ -315,6 +316,30 @@ async def combined_filter(
...
+                    # agent reads its MCP config at startup, and the bridge
+                    # proxy starts asynchronously. Launching early yields an
+                    # agent with no bridged tools and NO error, whose output is
+                    # then scored as a valid trajectory. Raises if unreachable.
+                    _http_mcp_configs = [
+                        c
+                        for c in bridge.mcp_server_configs
+                        if isinstance(c, MCPServerConfigHTTP)
+                    ]
+                    if _http_mcp_configs:
```
craig-meridian: [P2] Gate the centaur launch path too

The readiness wait is nested exclusively under the non-centaur branch, so centaur mode exposes the CLI alias before bridged MCP readiness has been verified. Kimi is especially affected because, as the new comment notes, it has no client-side blocking mechanism; invoking the alias immediately can therefore reproduce the same toolless first-call race.

The same nesting occurs in Claude Code, Codex, Gemini CLI, and OpenCode. Please perform an initial gate before the `if centaur` branch while retaining the per-attempt gate for unattended retries, and add coverage that distinguishes the two branches.

src/inspect_swe/_util/mcp_ready.py
```diff
@@ -0,0 +1,190 @@
...
+    False. Proceeding past an endpoint that serves no tools means launching an
+    agent that will silently have none, so callers should have a specific
+    reason to pass ``required=False``.
+    """
+    from inspect_ai.util import sandbox as sandbox_env
+
+    if not configs:
+        return True
+
+    sbox = sandbox_env()
```
craig-meridian: [P2] Probe the sandbox selected by the agent

This resolves `sandbox_env()` with no name, but every caller can be configured with `sandbox=` and creates the bridge/CLI in `sandbox_env(sandbox)`. In an Inspect task with multiple environments, calling `sandbox()` with `None` returns the configured default environment; it does not infer the explicitly selected agent sandbox. The probe therefore runs in the wrong container, typically waits 120 seconds and errors even though the named sandbox's endpoint is ready (or can accidentally validate an unrelated listener on the same port). Please pass the already-resolved `SandboxEnvironment` (or the sandbox name) into this helper and execute the probe there.

Relatedly, the `bridge` parameter is currently unused — nothing in the function body references it. Passing the resolved sandbox in (or in its place) would make the signature honest; the empty-server comment below also gives `bridge` a real use via `bridge.bridged_tools` if it stays.

src/inspect_swe/_util/mcp_ready.py
```diff
@@ -0,0 +1,190 @@
...
+                    "Accept: application/json, text/event-stream",
+                    "-X",
+                    "POST",
+                    "--data-binary",
+                    "@-",
+                    url,
+                ],
+                input=request,
+            )
+            tools = _tools_from_probe(result.stdout)
```
craig-meridian, line 277: [P2] Fail fast when the probe executable cannot run

Probe execution failures are currently treated exactly like an endpoint that is not ready: stderr and the return code are discarded, and the loop retries for the full 120 seconds before blaming the bridge. In particular, a custom sandbox image without `curl` cannot use bridged tools even when the bridge is healthy, but every sample waits two minutes and receives a misleading error.

Please validate the probe dependency once, or distinguish command-not-found/exec failures from expected connection failures, and raise immediately with stderr and the return code. The bridged-tools documentation should also state any sandbox dependency this introduces.

src/inspect_swe/_util/mcp_ready.py
```diff
@@ -0,0 +1,190 @@
...
+            if tools is None:
+                last_seen[name] = "no tool listing yet"
+                continue
+            environment_tools = [t for t in tools if t not in _NON_ENVIRONMENT_TOOLS]
```
craig-meridian: [P2] Do not wait for tools that the bridge does not contain

`BridgedToolsSpec` accepts an empty `tools` sequence, and the bridge exposes the registered tool mapping through `bridge.bridged_tools`. For an empty bridged server this loop can never succeed, so a valid configuration now waits 120 seconds and hard-errors even though zero tools is exactly what was configured.

Please use the bridge registry to skip empty servers—or validate them explicitly before polling—and test that case. The same registry can define the expected tool names; `WaitForMcpServers` is a client-side placeholder and cannot be returned by this proxy's `tools/list`, so the special-case filter is misleading. (Consulting the registry would also give the currently-unused `bridge` parameter its first actual use in this function.)

src/inspect_swe/_codex_cli/config.py
```diff
@@ -157,3 +159,23 @@ def check_codex_auto_review_version(version: str | None) -> None:
...
             "version) to codex_cli()."
         )
+
+
+def codex_mcp_server_config(
+    mcp_server: MCPServerConfig, bridged_server_names: AbstractSet[str]
+) -> dict[str, Any]:
+    """TOML table for one `mcp_servers.<name>` entry in codex config.
+
+    Bridged servers are marked `required = true`: codex >= 0.119.0 then blocks
```
craig-meridian: [P3] Correct the codex version boundary in this docstring too

Same correction as requested for the PR body: official release sources show `required` first present in 0.99.0 (`codex-rs/core/src/config/types.rs` at `rust-v0.99.0`, whose doc comment already carries the "`codex exec` exits with an error" semantics) and absent in 0.98.0. The raw MCP config struct's `deny_unknown_fields` there is schemars-only (schema generation), not serde, so older versions ignore the key exactly as this docstring says.

Please change `codex >= 0.119.0` to `codex >= 0.99.0` so the in-code record matches once the PR body is fixed.

craig-meridian (changes requested):
Thanks Sami/Claude! Review mostly by Claude & Codex with my oversight.

Before merging, please also update the PR body. It currently says the inherited bodyless `curl -sf -X POST` probe was valid and left unchanged, but the implementation now sends `tools/list` and requires a non-empty result. This repository uses the PR body as the squash-merge commit body, so the contradiction would be preserved in commit history.

While updating it, please also correct the Codex compatibility boundary: official release sources show `required` present in 0.99.0 and absent in 0.98.0; older raw MCP config structs ignore unknown fields. The same stale `>= 0.119.0` boundary is in the `codex_mcp_server_config` docstring in `_codex_cli/config.py` — flagged inline so it gets fixed in code too.

sjawhar:
Round-2 review addressed in [`0d78004d`](https://github.com/sjawhar/inspect_swe/commit/0d78004d8164ba464068f1e31a5af26e437cf6d0). Per-comment:

- **`mcp_ready.py:130` — recompute the remaining deadline before each endpoint probe.** `remaining` is now recomputed inside the `for name, (url, expected) in list(pending.items())` loop before every probe and breaks out as soon as it hits zero, so a slow first endpoint no longer eats the deadline of the ones behind it in the same pass. `sbox.exec` is bounded with the exact remaining budget (`timeout=max(1, int(probe_budget) + 1)`) so a hung sandbox transport cannot outlive the wall-clock deadline. Covered by `test_remaining_deadline_is_recomputed_per_endpoint` with two never-ready endpoints (interval-counting semantics needed ~400 probes on the input, per-probe semantics stops at ~4).

- **`kimi_code.py:329` — gate the centaur launch path too, and unnest across agents.** Every self-launching agent (claude_code, codex_cli, kimi_code, gemini_cli, opencode) now performs an initial `await wait_for_mcp_endpoints(...)` BEFORE the `if centaur:` branch and retains the per-attempt gate inside the retry loop for unattended-retry cases. Covered by `test_every_self_launching_agent_gates_before_the_centaur_branch` — AST-level guard that finds each agent's `_run_*_centaur` dispatch and asserts an awaited `wait_for_mcp_endpoints` at an earlier lineno, so a future refactor cannot silently re-nest the gate under `else`.

- **`mcp_ready.py:116` — probe the sandbox selected by the agent.** `wait_for_mcp_endpoints` gained a keyword `sandbox: SandboxEnvironment | str | None = None` that is passed through to `sandbox_env(...)` for resolution (or used directly if already resolved). Every call site now passes the same `sandbox=` the caller uses to build the bridge and launch the CLI. Covered by `test_sandbox_argument_selects_the_agent_environment`, which patches `inspect_ai.util.sandbox` and asserts it was resolved with the caller's name rather than `None`. The unused `bridge` parameter also gets its first real use in this PR: reading `bridge.bridged_tools` to decide which servers to probe and what tools to expect (see the next comment).

- **`mcp_ready.py:157` — fail fast when the probe executable cannot run.** A new `MCPProbeExecutableMissingError` is raised on the first probe when `curl` isn't in the sandbox: matched by exit code 127 or by `curl: not found`/`command not found`/`No such file or directory` on stderr. Connection failures (curl exit 6, 7, 28) stay in the "endpoint not ready yet" bucket and continue to retry. Covered by `test_missing_probe_executable_fails_fast_with_stderr` (exit 127) and `test_missing_probe_executable_matches_exec_error_message` (exit 1 with exec-time message), both of which assert `sbox.exec.await_count == 1` — no wait-then-time-out. The `MCPProbeExecutableMissingError` docstring records the sandbox dependency for image maintainers.

- **`mcp_ready.py:161` — do not wait for tools the bridge does not contain.** The gate now reads `bridge.bridged_tools` as the source of truth for both which servers to probe and what tools to expect. Servers registered with zero tools are skipped before the first probe (a valid configuration the old code polled forever); each probed server succeeds when the response contains at least one tool from `bridge.bridged_tools[<server>]`. This also lets the earlier `_NON_ENVIRONMENT_TOOLS` filter for `WaitForMcpServers` go — that placeholder cannot appear in a bridge-proxy `tools/list` response at all, so filtering it was misleading. Covered by `test_zero_tool_bridged_server_is_skipped_not_probed`, `test_listing_without_any_expected_tool_is_not_ready`, and `test_bridged_tools_registry_gates_expected_names`.

- **`_codex_cli/config.py:169` and PR body — codex version boundary.** Corrected to `codex >= 0.99.0` in both places. `required` first appears in `codex-rs/core/src/config/types.rs` at `rust-v0.99.0` (whose doc comment already carries the "`codex exec` exits with an error" semantics) and is absent in 0.98.0; the raw MCP config struct's `deny_unknown_fields` at that point is schemars-only (schema generation), not serde, so older versions ignore the key as this docstring already says.

The PR body is updated to describe the `tools/list` + expected-tools probe and the corrected codex version. `make check` clean (ruff format, ruff check, `mypy --strict` across 114 files, 0 issues); `pytest tests/test_mcp_ready.py`: 21 passed; full suite: 193 passed, 82 skipped, 1 failure in `test_k8s.py::test_k8s` (pre-existing `anthropic` SDK version mismatch unrelated to this PR).

This is Claude, on Sami's behalf.

src/inspect_swe/_util/mcp_ready.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+class MCPProbeExecutableMissingError(RuntimeError):
+    """The probe executable (``curl``) is missing from the sandbox image.
+
+    Raised immediately rather than after the full timeout. Without ``curl``,
+    the gate has no way to check readiness, and every sample would otherwise
+    wait the full ``timeout`` before failing with a misleading "endpoints
+    never answered" message. The sandbox image needs ``curl`` for this gate
+    to run; bridged-tools documentation notes the dependency.
```
craig-meridian, line 62: [P3] Add the curl prerequisite to the user-facing bridged-tools docs

This sentence says the bridged-tools documentation notes the dependency, but no docs changed in this PR: `docs/_snippets/agent_bridged_tools.md` (included by all six agent pages) doesn't mention `curl`. Users building custom sandbox images won't learn that bridged-tools readiness now requires it until execution fails. Please add a short prerequisite note there — which also makes this sentence true. (The Dockerfile example in `agent_mcp_servers.md` installs curl only incidentally, for the nodesource setup, so it doesn't cover this.)

src/inspect_swe/_claude_code/claude_code.py
```diff
@@ -434,6 +452,21 @@ async def execute(state: AgentState) -> AgentState:
                         # tool_result), so SpanBegin/End stay balanced.
                         consumer.reset()
 
+                        # Per-attempt gate: on unattended retries after this
+                        # loop restarts with --resume, the bridge could have
+                        # been restarted or briefly lost tool visibility. The
+                        # pre-centaur gate above only covered cold start;
+                        # this call is a no-op when endpoints are still ready.
```
craig-meridian: [nit — optional] First unattended launch probes every endpoint twice; "no-op" oversells the second call

My round-2 ask produced this shape (initial gate before the centaur branch, per-attempt gate retained), so this is a refinement, not a defect: the pre-centaur gate runs unconditionally and this first loop iteration immediately gates again. The helper has no readiness cache, so the second call is a real curl round trip per bridged server rather than a no-op — same in all five CLIs. The cost is one warm probe per server per launch, so I'm fine merging as is. If you want to tidy it: skip this gate on the first attempt (it's only needed when the loop restarts), or just reword "no-op" to something like "a cheap single-pass confirm".

src/inspect_swe/_util/mcp_ready.py
```diff
@@ -0,0 +1,303 @@
...
+                    "POST",
+                    "--data-binary",
+                    "@-",
+                    url,
+                ],
+                input=request,
+                # Bound sbox.exec itself with the exact remaining budget so
+                # a hung sandbox transport cannot outlive the deadline. +1
+                # gives curl's own --max-time room to fire first and report.
+                timeout=max(1, int(probe_budget) + 1),
```
craig-meridian: [nit — optional] Exec timeouts escape the gate's error semantics

`SandboxEnvironment.exec` raises `TimeoutError` when `timeout=` expires — and only after `timeout_retry` (default `True`) retries the command. Two edge-case consequences when the sandbox transport wedges past curl's own `--max-time`: the gate surfaces a bare `TimeoutError` instead of `MCPEndpointsUnreachableError` (bypassing `required=False` and losing the explanatory message), and the internal retries can stretch a single probe to roughly 3x its budget before the deadline logic sees anything. Wrapping this exec in `try/except TimeoutError` (treat as a failed probe and let the deadline decide) plus `timeout_retry=False` — the poll loop is already the retry mechanism — would keep the gate in charge. Only reachable when the transport itself hangs, so fine to defer or fix in a follow-up.

craig-meridian (commented):
Thank you Sami — round 2 addressed all six points, and I verified each in code and ran the branch locally (full suite green, ruff and mypy clean): the per-probe deadline recompute with the exec bounded, sandbox threading through every call site including ACP and antigravity, the registry-driven readiness check with the zero-tool skip, the curl fast-fail, the pre-centaur gates with their AST guard, the 0.99.0 docstring, and the PR body now matches the implementation.

What's left is one small documentation ask ([P3]) and two optional nits. If you choose not to take the optional ones, let me know so I know when I can merge this. Thanks!

sjawhar:
Round-3 in [`e21db6a0`](https://github.com/sjawhar/inspect_swe/commit/e21db6a0de48d67d10ab0fc75c8ff3dbb1f21f04): the P3 doc note plus both optional nits taken.

- **[P3] `agent_bridged_tools.md` — curl prerequisite.** Added a "Sandbox prerequisite" section spelling out that the readiness gate probes with `curl` inside the sandbox, so a minimal or distroless custom image without it now errors on the first probe (`MCPProbeExecutableMissingError`) instead of waiting out the 120s timeout, and pointing at `apt-get`/`apk`/etc. as the fix. The snippet is included by all six agent doc pages via `_snippets`, so every agent page picks it up.

- **[nit] First unattended launch probes every endpoint twice — taken.** The retry-loop `wait_for_mcp_endpoints` in every self-launching agent now fires only on iterations after the first, so the cold-start pre-centaur gate is not paid for twice per fresh launch. Guard matches each agent's existing "this is a retry" definition: `attempt_count > 0` (kimi_code, gemini_cli, opencode), extended with `cp.attempt == "resume"` for codex_cli, and `uncaught_error_count > 0` for claude_code (both are checkpoint-restore-safe because on restore the pre-centaur gate re-runs at function entry).

- **[nit] Exec timeouts escape the gate's error semantics — taken.** `sbox.exec` is now wrapped in `try/except TimeoutError` and called with `timeout_retry=False`. A wedged transport that outlasts `--max-time` is treated as a failed probe (recorded in `last_seen`, poll loop decides when to give up), and the sandbox transport's internal retries can no longer stretch one probe to ~3x its budget before the deadline logic sees anything. `test_exec_timeout_is_treated_as_a_failed_probe_not_a_bare_TimeoutError` pins both — `TimeoutError` converts to `MCPEndpointsUnreachableError` with `probe exec timed out` in the message, and every `sbox.exec` call carries `timeout_retry=False`.

`make check` clean (ruff format, ruff check, `mypy --strict` across 80 source files); `pytest tests/test_mcp_ready.py`: 22 passed. Ready when you are — this should be everything you flagged.

This is Claude, on Sami's behalf.

docs/_snippets/agent_bridged_tools.md
```diff
@@ -45,3 +45,7 @@ def investigator() -> Task:
 The `name` field identifies the MCP server and will be visible to the agent as a tool prefix. You can specify multiple `BridgedToolsSpec` instances to create separate MCP servers for different tool groups.
 
 See the [Bridged Tools](https://inspect.aisi.org.uk/agent-bridge.html#bridged-tools) documentation for more details on the architecture and how tool execution flows between host and sandbox.
+
+### Sandbox prerequisite
+
+The bridged-tools readiness gate probes each bridged MCP endpoint from inside the sandbox with `curl`, so the sandbox image must have `curl` on `PATH`. Debian/Ubuntu, Alpine, and Fedora base images ship it by default; a minimal or `distroless` custom image without it will error immediately on agent launch with `MCPProbeExecutableMissingError`, rather than after the 120s readiness timeout. Add `curl` to the image (`apt-get install -y curl`, `apk add curl`, or equivalent) if you build your own.
```
craig-meridian, line 51: [P3] Don't claim the distro base images include curl by default

The prerequisite is now documented in the right place, but 'Debian/Ubuntu, Alpine, and Fedora base images ship it by default' is wrong for three of the four. The official `debian` and `ubuntu` images don't include `curl` at all (verified against `debian:latest`, `debian:stable-slim`, and `ubuntu:latest`), and the official Alpine image is intentionally minimal — busybox `wget` only; `curl` requires `apk add`. Only `fedora` ships it. Full `python:3.x` images do have it (via `buildpack-deps`), which may be where the impression came from — but `python:3.x-slim` does not.

As written, the sentence falsely reassures exactly the users this note exists to warn: someone on an Alpine or Debian base reads 'ships it by default', skips the install, and hits `MCPProbeExecutableMissingError` anyway.

Please replace the clause with something like: 'Full `python:3.x` and `fedora` images include it, but the official `debian`, `ubuntu`, and `alpine` base images (and slim/distroless variants) do not.' The installation guidance that follows is good.

craig-meridian (commented):
Thanks Sami, just one correctness point in docs to address. Any other findings from AI agents were nits, which I'm not pursuing at this stage in the multi-round review process.

sjawhar:
Two additions pushed, folding the rest of the fork's MCP-readiness work into this PR so it ships as one story (per Sami):

- [`43464b94`](https://github.com/sjawhar/inspect_swe/commit/43464b94) **feat: make MCP readiness timeout configurable on every gated agent.** Threads an `mcp_ready_timeout` option through claude_code, codex_cli, gemini_cli, kimi_code, and antigravity into both readiness gates (initial pre-centaur and retry-loop). Motivation: a 720-sample eval-set died end-to-end on the fixed 120s default — the bridge answers in ~1s at one sample, but not while ~400 sandboxes share the runner's event loop. Guarded by an AST test asserting every agent forwards the option to every `wait_for_mcp_endpoints` call.
- [`3a2571f9`](https://github.com/sjawhar/inspect_swe/commit/3a2571f9) **fix(codex): give MCP servers a startup timeout that covers a slow sandbox.** Codex bounds its session-start MCP wait with a per-server `startup_timeout_sec` nothing set, so a slow-starting sandboxed server failed silently and the agent ran toolless (2.4% of 250 samples at 150-way concurrency; 87/4130 in a production collection). `setdefault`, so an explicit author value still wins.

Both are the same failure family this PR closes (agent reaches the model without its tools). `make check` clean, `pytest tests/test_mcp_ready.py tests/test_codex_config.py tests/test_claude_code_env.py`: 61 passed. Understood this resets your round-3 sign-off — sorry for the moving target, and happy to split these back out if you'd rather merge the reviewed scope as-is.

This is Claude, on Sami's behalf.

craig-meridian (changes requested):
Thanks Sami. Since you called these additions out and offered to split them: although they are related to MCP readiness, they introduce two new behaviours after this PR had converged — a public timeout option across the agent APIs and a Codex-specific startup-timeout policy. Please move commits `43464b9` and `3a2571f` into a focused follow-up PR, and keep #106 to the previously reviewed readiness work, the outstanding curl wording correction, and resolving the conflict with main.

To save another review round, please carry these findings into the follow-up:

- Add `mcp_ready_timeout` to OpenCode and the public ACP path, and make the coverage include every gated entry point rather than a hand-picked subset.
- Apply the Codex 300-second startup default only to bridged servers; cover bridged, static, and explicit-override cases in the generated configuration tests.
- Add `mcp_ready_timeout` to the user-facing option tables for every agent that exposes it.

Before #106 merges:

- Complete the curl prerequisite wording correction already requested in the previous review (this is not a new request): full `python:3.x` and Fedora images include it, while the official Debian, Ubuntu, Alpine, slim, and distroless images generally do not.

sjawhar:
Claude here, on Sami’s behalf. Split done as requested: this PR now retains the reviewed readiness work and curl wording correction. The two requested commits moved to focused follow-up PR #123.

craig-meridian (approved):
Thanks for this Sami, approved!

PR #107, opened 2026-08-07
fix(codex_cli): support Codex Multi-Agent V2 — package-archive install and sub-agent span reconstruction

dragonstyle:
Two fixes that together make Codex Multi-Agent V2 (codex >= 0.146, forced for gpt-5.6 models) work end to end. Both were verified against a live gpt-5.6-sol multi-agent eval (codex 0.147.0, docker sandbox).

### Problem

`codex_cli()` installs only the single `codex` binary. Codex releases since rust-v0.133.0 ship companion executables the CLI resolves at runtime — most critically `codex-code-mode-host`, which Code Mode runs in as a standalone host process (`features.code_mode_host` is stable and default-on in 0.147.0).

With codex >= 0.146 and gpt-5.6+ models, Code Mode is the execution path for Multi-Agent V2 sub-agents. Missing the host binary, Code Mode fails closed:

```
warning: Code Mode is unavailable because failed to spawn code-mode host
/var/tmp/.5c95f967ca830048/codex-code-mode-host: host executable was not found.
Code mode will fail closed; enable `features.code_mode_host` and install `codex-code-mode-host`.
```

Sub-agents then cannot execute anything ("the shared execution host is missing") and every multi-agent task fails.

### Fix

Install the `codex-package-<target>.tar.gz` release asset instead of the single binary. It is codex's canonical bundle:

```
bin/codex
bin/codex-code-mode-host
codex-path/rg
codex-resources/bwrap
codex-resources/zsh/...
```

- **Asset-presence fallback:** the package asset first shipped with rust-v0.133.0; when a pinned older release doesn't have it, resolution falls back to the single `codex-<target>.tar.gz` and the install path is unchanged.
- **Verbatim caching:** package archives are cached as downloaded, so the GitHub asset digest stays verifiable on cache reads (the single-binary path can't verify cached artifacts because `post_download` transforms them).
- **In-sandbox extraction:** the archive is written into the sandbox and extracted into a versioned directory (`$INSTALL_DIR/codex-<version>-<platform>/`), returning `bin/codex` as the agent binary. Running from the package layout is what lets codex's `CodexPackageLayout::from_exe` discover its bundled resources. A `test -x` short-circuit skips the write + extract when that version is already installed.
- Non-package agents (claude_code, etc.) are unchanged.
- Also: `cached_agent_binaries` name parsing handles the new artifact shape, and `-models.json` catalog caches are excluded from the codex binary-cache glob so cleanup can't count them against real binaries.

### Problem

Multi-Agent V2 changed every signal the `CodexConsumer` used to build the agent-span tree, so sub-agent spans were named the generic "agent" and every model event landed flat on the outer `codex_cli` span (0 events attributed).

### Fix

- `spawn_agent` calls carry `task_name` (not `agent_type`) and an encrypted message, so prompt-substring attribution matched nothing → name spans after `task_name` and attribute calls by their inbound `agent_message` recipient path (preserved on `ContentText.internal` by the agent bridge in inspect_ai >= 0.3.253), with the V1 substring heuristic as fallback.
- `spawn_agent` results carry `{"task_name": "/root/<name>"}` → accept the absolute task path as the thread id.
- Completion arrives as a `FINAL_ANSWER` agent_message from the child (the `wait_agent` result no longer carries per-thread status) → close the author's span on `FINAL_ANSWER`.
- V1 behavior unchanged (regression-guarded in tests).

## Dependencies

Full gpt-5.6 multi-agent support also requires the inspect_ai agent bridge fix for `agent_message` items (UKGovernmentBEIS/inspect_ai#4787): without it the bridge rejects V2 requests outright, and span attribution by recipient (fix 2) degrades to the V1 fallback. This PR is safe to merge independently.

## Testing

- Package install: 8 unit tests — package-asset preference, single-binary fallback, verbatim caching + checksum-verified cache hits, in-sandbox extraction, already-installed skip, missing-entrypoint guard, catalog exclusion (`tests/test_agentbinary_install.py`, `tests/test_codex_agentbinary.py`).
- Spans: V2 reconstruction + V1 regression tests (`tests/test_codex_subagent_spans.py`).
- ruff + mypy clean; CI green.
- Verified end to end on a gpt-5.6-sol multi-agent eval that fails on main: with this branch the package extracts in-sandbox, no Code Mode warning, both spawned sub-agents execute and the task scores correct; spans are named `write_fizzbuzz`/`write_primes` with a nested grandchild, 89/113 model events attributed to sub-agent spans (previously 0), and all spans closed.

## Merge note

The branch intentionally carries two `fix:` commits — please use merge-commit or rebase (not squash) so Release Please records both changelog entries.

### DIFF
--- a/src/inspect_swe/_codex_cli/_events/detection.py
+++ b/src/inspect_swe/_codex_cli/_events/detection.py
@@ -38,6 +38,52 @@ class SpawnedAgent:
     agent_type: str
     message: str
     reasoning_effort: str | None
+    task_name: str | None = None
+    """Multi-Agent V2 task name (e.g. "write_fizzbuzz"); None under V1."""
+
+    @property
+    def name(self) -> str:
+        """Display name for the agent's span (V2 task_name, else V1 agent_type)."""
+        return self.task_name or self.agent_type
+
+
+def agent_message_recipients(input_messages: list[ChatMessage]) -> set[str]:
+    """Recipients of the agent_message items in a request's input.
+
+    Multi-Agent V2 delivers inter-agent messages as `agent_message` input items;
+    the bridge preserves each raw item (author/recipient) on ContentText.internal.
+    Every agent_message in a request is inbound to the requester, so the
+    recipient path (e.g. "/root/write_fizzbuzz") identifies the calling agent.
+    """
+    return {
+        recipient
+        for item in _agent_message_items(input_messages)
+        if isinstance(recipient := item.get("recipient"), str) and recipient
+    }
+
+
+def final_answer_authors(input_messages: list[ChatMessage]) -> set[str]:
+    """Authors of FINAL_ANSWER agent_message items in a request's input.
+
+    A FINAL_ANSWER is a sub-agent's terminal return under Multi-Agent V2 (the
+    `wait_agent` result no longer carries per-thread completion status), so its
+    author path marks that agent's thread as completed.
+    """
+    authors: set[str] = set()
+    for item in _agent_message_items(input_messages):
+        author = item.get("author")
+        if not (isinstance(author, str) and author):
+            continue
+        for part in item.get("content") or []:
+            if (
+                isinstance(part, dict)
+                and part.get("type") == "input_text"
+                and isinstance(text := part.get("text"), str)
+                and text.lstrip().startswith("Message Type: FINAL_ANSWER")
+            ):
+                authors.add(author)
+                break
+    return authors
 
 
 def find_spawned_agents(tool_calls: list[ToolCall] | None) -> list[SpawnedAgent]:
@@ -51,12 +97,14 @@ def find_spawned_agents(tool_calls: list[ToolCall] | None) -> list[SpawnedAgent]
         if not isinstance(message, str) or not message:
             continue
         reasoning = args.get("reasoning_effort")
+        task_name = args.get("task_name")
         result.append(
             SpawnedAgent(
                 call_id=tc.id,
                 agent_type=str(args.get("agent_type") or "agent"),
                 message=message,
                 reasoning_effort=str(reasoning) if reasoning else None,
+                task_name=str(task_name) if task_name else None,
             )
         )
     return result
@@ -88,12 +136,15 @@ def spawn_result(message: ChatMessageTool) -> SpawnResult | None:
     The result is correlated to its spawn call by `message.tool_call_id`, so the
     caller can bind thread_id → span without any ordering assumptions. The
     `nickname` (Codex's friendly per-agent name) is surfaced for tool views.
+
+    Multi-Agent V2 results carry `{"task_name": "/root/<name>"}` instead of
+    `agent_id`/`nickname`; the absolute task path serves as the thread id.
     """
     if message.function != SPAWN_AGENT:
         return None
     data = _loads(message.text)
     if isinstance(data, dict):
-        agent_id = data.get("agent_id")
+        agent_id = data.get("agent_id") or data.get("task_name")
         if isinstance(agent_id, str) and agent_id:
             nickname = data.get("nickname")
             return SpawnResult(
@@ -133,6 +184,21 @@ def is_compaction_request(input_messages: list[ChatMessage]) -> bool:
 # ---------------------------------------------------------------------------
 # internal
 # ---------------------------------------------------------------------------
+
+
+def _agent_message_items(input_messages: list[ChatMessage]) -> list[dict[str, Any]]:
+    """Raw agent_message items stashed on user-message content by the bridge."""
+    items: list[dict[str, Any]] = []
+    for msg in input_messages:
+        if not isinstance(msg, ChatMessageUser) or isinstance(msg.content, str):
+            continue
+        for content in msg.content:
+            internal = getattr(content, "internal", None)
+            if isinstance(internal, dict):
+                item = internal.get("agent_message")
+                if isinstance(item, dict):
+                    items.append(item)
+    return items
 
 
 def _loads(text: str) -> Any:
--- a/src/inspect_swe/_codex_cli/agentbinary.py
+++ b/src/inspect_swe/_codex_cli/agentbinary.py
@@ -30,15 +30,25 @@ def codex_cli_binary_source() -> AgentBinarySource:
         # Get release information
         release = await _fetch_release_assets(version)
 
-        # Get the platform-specific asset
+        # Get the platform-specific asset. Prefer the full package archive
+        # (ships companion executables codex needs at runtime, e.g.
+        # codex-code-mode-host, which gpt-5.6+ models require for execution;
+        # first published with rust-v0.133.0) and fall back to the single
+        # codex binary for older releases.
         arch = _platform_to_codex_arch(platform)
-        asset_name = f"codex-{arch}.tar.gz"
+        candidates = [
+            (f"codex-package-{arch}.tar.gz", True),
+            (f"codex-{arch}.tar.gz", False),
+        ]
 
         # Find the matching asset
+        assets = {a["name"]: a for a in release.get("assets", [])}
         asset = None
-        for a in release.get("assets", []):
-            if a["name"] == asset_name:
-                asset = a
+        package = False
+        for asset_name, is_package in candidates:
+            asset = assets.get(asset_name)
+            if asset is not None:
+                package = is_package
                 break
 
         if asset is None:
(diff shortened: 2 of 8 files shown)


src/inspect_swe/_util/agentbinary.py
```diff
@@ -82,31 +91,60 @@ async def ensure_agent_binary_installed(
...
-            binary_bytes: bytes | None = read_cached_binary(
-                source, version, platform, None
-            )
+            if source.cached_package_path is not None:
+                binary_bytes = read_cached_file(
+                    source.cached_package_path(version, platform), None
+                )
+                package = binary_bytes is not None
+            if binary_bytes is None:
+                binary_bytes = read_cached_binary(source, version, platform, None)
```
dragonstyle, line 110: **Stale single-binary cache silently defeats the package install for pinned versions.** A user upgrading inspect_swe who previously downloaded, say, codex 0.147.0 will have `codex-0.147.0-<platform>` cached but no package archive — so this single-binary fallback wins, `package` stays `False`, and the exact Code Mode failure this PR fixes comes back silently. That's the population most likely to hit the bug (existing users, warm caches, pinned versions).

Options, roughly in order of preference:
- When `cached_package_path` is defined and only the single-binary cache hits, still call `resolve_version` (network permitting) to check for a package asset, falling back to the cached binary offline.
- Or encode that packages exist from 0.133.0 (e.g. a `package_min_version` field on the source) and skip the single-binary short-circuit for those versions.
- At minimum, `trace()` a warning when a pinned version ≥ 0.133.0 installs from the single-binary cache.

A test asserting that a pinned version with a stale single-binary cache still ends up with the package would lock in whichever fix you choose.

src/inspect_swe/_util/agentbinary.py
```diff
@@ -82,31 +91,60 @@ async def ensure_agent_binary_installed(
...
-        await sandbox_exec(sandbox, f"chmod +x {binary_path}", user="root")
+        if package:
+            if source.package_entrypoint is None:
+                raise RuntimeError(
+                    f"{source.agent} resolved a package archive but the source "
+                    "does not define a package_entrypoint"
+                )
+            binary_path = f"{install_path}/{source.package_entrypoint}"
+            # skip write + extract if this version's package is already installed
+            probe = await sandbox.exec(bash_command(f"test -x {binary_path}"))
```
dragonstyle: The probe runs as the default sandbox user while extraction runs as root. If the install dir ends up non-traversable for the default user, `test -x` fails on every call and the archive is re-written and re-extracted each time — correct but wasteful. Consider running the probe as root to match the extraction.

src/inspect_swe/_util/agentbinary.py
```diff
@@ -82,31 +91,60 @@ async def ensure_agent_binary_installed(
...
+            if not probe.success:
+                archive_path = f"{install_path}.tar.gz"
+                await sandbox.write_file(archive_path, binary_bytes)
+                await sandbox_exec(
+                    sandbox,
+                    f"mkdir -p {install_path} && "
+                    f"tar -xzf {archive_path} -C {install_path} && "
+                    f"rm -f {archive_path} && "
+                    f"chmod +x {binary_path}",
```
dragonstyle, line 166: Only `bin/codex` gets `chmod +x` — the companion executables (`codex-code-mode-host`, `codex-resources/bwrap`) rely on tar preserving mode bits. That's normally true (and the live eval confirms it), just noting the assumption is load-bearing.

src/inspect_swe/_util/agentbinary.py
```diff
@@ -134,31 +172,44 @@ async def download_agent_binary_async(
         resolved = await source.resolve_version(version, platform)
         with _resolve_version_lock:
             _resolved_versions[cache_key] = resolved
-    version, expected_checksum, download_url = resolved
+    version, expected_checksum, download_url = resolved[:3]
```
dragonstyle: Nit: attribute access (`resolved.version`, `resolved.expected_checksum`, `resolved.download_url`) would read better than slicing the NamedTuple.

src/inspect_swe/_codex_cli/agentbinary.py
```diff
@@ -55,13 +65,20 @@ async def resolve_version(
...
-        return AgentBinaryVersion(version, expected_checksum, download_url)
+        return AgentBinaryVersion(version, expected_checksum, download_url, package)
 
     def cached_binary_path(version: str, platform: SandboxPlatform) -> Path:
         return cached_binary_dir / f"codex-{version}-{platform}"
 
+    def cached_package_path(version: str, platform: SandboxPlatform) -> Path:
+        return cached_binary_dir / f"codex-package-{version}-{platform}.tar.gz"
+
     def list_cached_binaries() -> list[Path]:
```
dragonstyle, line 76: Eviction accounting changed shape: a version can now consume two slots (single binary + package archive) against `keep_count=3` in `_cleanup_binary_cache`, so fewer distinct versions may survive; `cached_agent_binaries()` also now lists two rows for the same version. Likely acceptable, but worth a deliberate call.

src/inspect_swe/_codex_cli/_events/detection.py
```diff
@@ -38,6 +38,52 @@ class SpawnedAgent:
...
+    for item in _agent_message_items(input_messages):
+        author = item.get("author")
+        if not (isinstance(author, str) and author):
+            continue
+        for part in item.get("content") or []:
+            if (
+                isinstance(part, dict)
+                and part.get("type") == "input_text"
+                and isinstance(text := part.get("text"), str)
+                and text.lstrip().startswith("Message Type: FINAL_ANSWER")
```
dragonstyle, line 82: Inherent fragility, acceptable but worth acknowledging: completion detection matches the literal string `"Message Type: FINAL_ANSWER"` in the envelope. A codex-side format change would leave spans open until `reset()` closes them (balanced, just less precise). The real-log fixtures make this a reasonable trade.

dragonstyle (commented):
Reviewed both fixes end to end (ran the 25 new/changed tests locally — all pass). The span-reconstruction logic is well-defended: attribution falls through to the outer span on any ambiguity, and `_close_thread` is idempotent against replayed `FINAL_ANSWER` messages, so misattribution degrades safely rather than landing on a wrong sub-agent. V1 behavior is regression-tested. No new security concerns: the checksum is integrity (not authenticity) protection as before, and the package path actually improves cache verifiability.

Inline comments below — one cache-interaction issue worth addressing (or explicitly accepting) before merge, the rest minor.

One process note: the merge note asks for merge-commit/rebase, but CLAUDE.md says the repo squash-merges and offers a mechanism for exactly this case — body lines starting with `<type>:` are parsed as extra changelog entries. You could squash-merge and add a line like `fix: reconstruct sub-agent spans under Codex Multi-Agent V2` to the PR body to get the second changelog entry without deviating from the squash convention. Worth deciding explicitly before merge.

tests/test_agentbinary_install.py
```diff
@@ -0,0 +1,178 @@
...
+import anyio
+import pytest
+from inspect_ai.util import SandboxEnvironment
+from inspect_swe._util import agentbinary
+from inspect_swe._util.agentbinary import (
+    AgentBinarySource,
+    AgentBinaryVersion,
+    download_agent_binary_async,
+    ensure_agent_binary_installed,
+)
```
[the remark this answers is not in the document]
dragonstyle: Fixed in 2d54080 (pushed just before this review posted) — the imports are re-sorted and `ruff check` passes repo-wide under the pinned ruff 0.12.11 (`make check` clean locally).

src/inspect_swe/_util/agentbinary.py
dragonstyle: Implemented option 1 in 5e91b79: package-capable sources now only honor a *package* cache hit directly for pinned versions; a single-binary cache entry no longer short-circuits resolution, so the package asset is discovered and installed (resolution results stay memoized, so it's one release-metadata fetch per process). When resolution fails, it falls back to the cached single binary with a trace message, preserving the offline guarantee. A downloaded package also deletes the superseded single-binary cache entry for the same version. Locked in by `test_stale_single_binary_cache_still_installs_package` and `test_offline_pinned_version_falls_back_to_cached_binary`.

src/inspect_swe/_util/agentbinary.py
dragonstyle: Fixed in 5e91b79 — the probe now runs as root to match the extraction.

src/inspect_swe/_util/agentbinary.py
dragonstyle: Added a code comment in 5e91b79 documenting that tar mode-preservation is load-bearing for the companion executables (verified against the live 0.147.0 archive: bin/*, codex-resources/* all carry executable bits).

src/inspect_swe/_util/agentbinary.py
dragonstyle: Fixed in 5e91b79 — attribute access instead of slicing.

src/inspect_swe/_codex_cli/agentbinary.py
dragonstyle: Deliberate call, with the sharp edge removed in 5e91b79: a package download now deletes the superseded single-binary entry for the same version, so the two-slots-per-version case only persists for genuinely mixed-era caches (a pre-0.133 pin alongside package-era versions), where both artifacts are real and should each hold a slot. `cached_agent_binaries()` listing both rows in that case reads as accurate rather than confusing.

src/inspect_swe/_codex_cli/_events/detection.py
dragonstyle: Agreed on the trade-off as stated: a format change degrades to spans closing at `reset()` (balanced, less precise) rather than anything worse, and the real-log fixtures will catch the drift when codex changes the envelope. Leaving as-is.

src/inspect_swe/_util/agentbinary.py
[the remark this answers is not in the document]
dragonstyle: Addressed in 5e91b79 with option (a): pinned versions on package-capable sources resolve before honoring a single-binary cache hit, falling back to the cached binary when resolution fails (offline guarantee preserved). The superseded single-binary cache entry is also deleted once the package is downloaded. See the sibling thread for details + tests.

dragonstyle:
@review

PR #111, opened 2026-08-08
feat: filter-time sub-agent attribution across all agents (agent bridge context)

dragonstyle:
## Why

Part of #131 (multi-agent delegation across bridged CLIs).

Bridged CLI agents now spawn sub-agents — Codex via Multi-Agent V2, Claude Code via its Task tool, OpenCode via its task tool — and every sub-agent's model calls flow through the same bridge as the top-level agent's. A `filter` passed to any inspect_swe agent therefore fires on *all* of them, with no way to tell whose request it is looking at.

Anyone doing mid-episode control from a filter breaks as soon as the CLI delegates: steering written for the top-level agent leaks into sub-agent conversations, turn counting mixes threads, and an early-stop ends a sub-agent's turn instead of the episode. The originating case was a matched-arms evaluation (same task under `react()` and under a bridged CLI, byte-identical steering in both) where delegation on the CLI side silently broke the comparison — and the only workaround was parsing another package's rendering internals out of the message history.

## What this adds

Every `inspect_swe` agent now classifies each bridged model request — is it the top-level agent's own thread (`root`), a delegated sub-agent's (`subagent`), CLI machinery like compaction or title generation (`utility`), or undeterminable (`unknown`) — **before** the caller's `filter` runs. A filter author confines control logic to the top-level agent with one check:

```python
from inspect_swe import is_root_agent

async def my_filter(model, messages, tools, tool_choice, config):
    if not is_root_agent():
        return None   # never steer a sub-agent or utility thread
    ...
```

`is_root_agent()` is the recommended, permissive gate: `False` only for requests attributed to `"subagent"`/`"utility"`; `True` for `"root"`, `"unknown"`, and no-context (so steering degrades gracefully rather than silently going dark when attribution can't be made). `current_agent_bridge_context()`, `AgentBridgeContext`, and the narrower `is_sub_agent()` are also re-exported from `inspect_swe`.

Built on **UKGovernmentBEIS/inspect_ai#4791**, which adds the minimal upstream surface this depends on: the `AgentBridgeContext` type, `current_agent_bridge_context()` / `set_agent_bridge_context()`, a default `"unknown"` stamp at bridge-request start, and `current_bridge_request()` (the raw, pre-alias-resolution requested model slug — the structural signal this PR's per-CLI resolvers read). This PR requires an `inspect_ai` release containing that PR (see the checklist below — UKGovernmentBEIS/inspect_ai#4791 is still open upstream as of 2026-09-03).

All per-CLI resolution lives in a new shared `_util/agentcontext.py` (`classify_filter` wrapper + `slug_map_classifier`/`static_root_classifier` helpers), installed at every `sandbox_agent_bridge` call site across all seven native agents plus the three ACP variants, consulted by the same resolver the existing `ModelEventSink` consumers already use for transcript span attribution — so the filter-time answer and the transcript's span attribution can never disagree.

## Per-Agent Guarantees

(copied from `docs/sub-agent-context.qmd`, verified accurate against Step-1 live validation below)

| Agent | Sub-agent detection | Utility detection | Notes |
|-------|---------------------|--------------------|-------|
| `claude_code` | Structural: the requested model carries the synthetic subagent slug (`CLAUDE_CODE_SUBAGENT_MODEL`), including custom agents defined via front-matter `model:`. Falls back to a pending-subagent prompt-substring match if the slug signal is missing. | Only when `haiku_model` is configured distinct from the presented model — otherwise the small-fast slug is indistinguishable from main-thread traffic. | A drift canary logs a warning (once) if Claude Code ever stops honoring the subagent slug this classification depends on. |
| `codex_cli` | Structural: tracks spawned sub-agent threads via `spawn_agent` tool calls and `agent_message` recipient routing. | The guardian (`auto_review`) model slug, plus Codex's own local-compaction request marker. | An unattributed call while sub-agents are open resolves to `"unknown"` rather than guessing "root". |
| `opencode` | Structural, but only for the three built-in (`general`/`explore`/`scout`) sub-agents — routed to a distinct sentinel model via injected per-agent config — and only when `opencode_model` is an `anthropic/` model, the one provider whose sentinel id is live-verified against OpenCode's model catalog (a catalog-rejected sentinel hard-fails the sub-agent's Task call, so unverified providers get no injection and classify `"root"`). Custom Task-tool agents aren't covered: one left at the default model inherits the primary slug and classifies `"root"`; one configured with its own `model:` gets a slug in neither map and classifies `"unknown"`. | Best-effort via `small_model` only (the per-agent utility overrides are not injected — not live-verified, and on an install where those names aren't reserved they would create spawnable agents). Live evidence so far shows title-gen under the primary slug, i.e. `"root"`. | Sentinel selection avoids colliding with the caller's own primary model slug; a collision omits the override so that role's traffic classifies `"root"` rather than injecting an id that could fail. |
| `gemini_cli` | None — gemini-cli's own subagent feature has no reserved model slug, so sub-agent traffic isn't distinguishable from root by slug alone. | Known internal utility calls (loop detection, web search/fetch, edit correction, next-speaker checks, some compaction variants) resolve to a small, known set of concrete model slugs. | Two compaction aliases resolve to the same slug as the presented model, so those specific calls are honestly under-attributed as `"root"`, not misclassified. |
| `kimi_code` | None — Kimi Code has no delegation feature. | Its own auto-compaction summarizer is matched by a fixed marker string carried in the request. | Everything else classifies `"root"`. |
| `mini_swe_agent` / `antigravity` | None — neither agent supports delegation. | None. | Every bridged request classifies `"root"`. |
| ACP variants (`interactive_claude_code`, `interactive_codex_cli`, `interactive_gemini_cli`) | Slug-only, via the same static root/subagent/utility slug maps their native counterparts use for their structural signal. | Slug-only, same maps. | No JSONL/event-stream consumer is available in the ACP path to drive the richer, stateful attribution the native agents layer on top — so ACP classification is exactly as good as the underlying agent's slug signal alone (e.g. `gemini_cli`'s "no subagent slug" gap, `codex_cli`'s guardian slug). `interactive_claude_code` presents sub-agents under a synthetic `<presented>-subagent` slug (aliased to the primary model when `subagent_model` is unset, as the native `claude_code` does), so Task sub-agent traffic is attributable out of the box; utility detection requires a `haiku_model` distinct from the presented/opus/sonnet models. |

`"unknown"` is a first-class, honest answer, not a bug — see "A Note on unknown" in `docs/sub-agent-context.qmd` for the full rationale.

## Validation Evidence

**Probes (pre-implementation, decided the matrix cells above):**
- **P1** (claude_code slug universality, live docker): with `CLAUDE_CODE_SUBAGENT_MODEL` set, ALL Task-tool subagents — including a custom agent with `model:` front-matter — carried the subagent slug as their raw requested model. Confirmed the structural signal is universal on CC 2.1.220.
- **P2** (opencode per-agent models, live docker): a project-level `opencode.json` subagent `model` override WAS honored end-to-end at the wire (`current_bridge_request().model` carried the configured id) — but only for real catalog model ids; a synthetic non-catalog slug was rejected outright by OpenCode's runtime model resolution (`"Model not found"`), which is why `opencode()`'s sentinels are real catalog ids aliased to the served model, not synthetic slugs like claude_code's.

**Step 1 end-to-end validation (this PR, live docker + live models, 5 of a 6-run budget used):**

| Run | Agent / config | Result |
|---|---|---|
| a | `claude_code`, **default** config (no `subagent_model`) | 6 requests: 2 `root` (presented slug `claude-opus-4-8`), 4 `subagent` (synthetic slug `claude-opus-4-8-subagent`). Zero `unknown`. No drift-canary warning logged. All assertions passed. |
| b | `codex_cli`, delegating prompt, steering-style filter (`if not is_root_agent(): return None`, plus logging) | 9 requests: 4 `subagent` (all gated `is_root_agent()==False`), 1 `root`, 4 `unknown` (all gated `is_root_agent()==True`) — the `unknown` results were calls made by the root thread carrying a V1 completion notification that `classify` did not yet process (review finding, fixed in 8d1bb5d: `classify` now closes completed threads before attributing, so these classify `root`). No subagent call was ever miscategorized as root or vice versa, and the gate value tracked `kind` exactly as designed (permissive: True for root+unknown, False only for subagent). *(First attempt against the default `anthropic/claude-opus-4-8` failed for an unrelated harness reason — Anthropic's API rejects Codex's `tool_choice.disable_parallel_tool_use=true` for programmatic tool calling — retried against `openai/gpt-5-mini`, Codex's native backend, which resolved it.)* |
| c, c2 | `opencode`, delegating prompt (general subagent, run twice) | 7 + 4 = 11 requests across both runs: all `general`-subagent delegations correctly classified `subagent` (sentinel slug `claude-haiku-4-5-20251001`) and all main-thread turns `root` (presented slug). Zero `unknown` in either run. Neither fresh run triggered a title/summary/compaction call at all (plausible async race between OpenCode's title-gen and the short session's teardown). Earlier live evidence captured during this same effort's development (`requests_run{A,B}.jsonl`, same repo) **did** capture a title-generation call, and it landed under the **primary** slug (`kind=="root"`), not the utility sentinel — confirming the config override is a silent no-op for title-gen, exactly the known limitation already documented in `opencode.py`'s module docstring and the guarantee table above. |

No CODE bugs found in Step 1 — every observed classification was either correct or the documented honest-`unknown`/no-op fallback.

**Step 2 (fast suite + gates):**
```
158 passed, 2 skipped in 4.33s
```
(`tests/test_agent_context_util.py tests/test_agent_context_static.py tests/test_agent_context_gemini.py tests/test_agent_context_opencode.py tests/test_claude_code_agent_context.py tests/test_claude_code_model.py tests/test_claude_code_exit.py tests/test_codex_subagent_spans.py tests/test_kimi_setup.py tests/test_opencode_setup.py tests/test_mini_swe_setup.py`)

`ruff check src tests` → **All checks passed!**
`mypy src/inspect_swe` → **Success: no issues found in 80 source files**

## Design Doc

Full design rationale, locked decisions, and empirical background facts: `plans/2026-08-08-agent-bridge-context-design.md` (local to this branch, intentionally untracked).

## Before Undrafting

- [ ] `inspect_ai` release containing UKGovernmentBEIS/inspect_ai#4791 published
- [ ] bump the `inspect_ai` floor in `pyproject.toml` + `uv lock` to that release
- [ ] CI green against the released `inspect_ai` — **CI will fail on the new tests until that release is out; this is expected and not a regression to chase down.**
- [x] follow-up issue filed for the pre-existing broken docker test: #112

## Tooling Disclosure & Agent Review

Each task in this plan was implemented by a Sonnet subagent, followed by a two-stage review before merge: a Sonnet spec-conformance pass, then a Fable final-arbiter pass, with a fix-first loop back to implementation on any confirmed finding — same process as inspect_ai#4791. Substantive catches from those review passes, by task:

- **kimi_code auto-compaction classifier**: the first pass matched Kimi's compaction marker against any message; review caught that this could mis-flip a main-thread request to `utility` if a tool result ever echoed the marker text (Kimi's own docs/source), and pushed the match down to "last message is a `ChatMessageUser` containing the marker" specifically — the false-positive direction is the harmful one (tool output can't flip a real request to `utility`; a missed match just reverts to `root`, which no gate treats differently). Also added a live-marker drift canary symmetric with claude_code's.
- **claude_code subagent slug collision set**: the initial design let an unset `subagent_model` inherit the presented model's name verbatim, which would collide with `presented`/`opus`/`sonnet`/`haiku` and make subagent traffic indistinguishable from whichever role it collided with. Review pushed for the explicit invariant in `resolve_claude_code_models`: `models.subagent` is never allowed to equal any of the other four role names — an unset (or coincidentally-colliding) `subagent_model` gets the synthetic `"<presented>-subagent"` name instead. Paired with a drift-canary warning if live Claude Code traffic ever stops carrying that slug.
- **opencode sentinel collision guard + config snapshot test**: review caught that a caller's own `opencode_model` could legitimately collide with a hardcoded sentinel default (e.g. explicitly setting `opencode_model` to the exact model this package would otherwise pick as its subagent sentinel), silently swallowing subagent/utility classification since `root_slugs` is checked first. Fixed with an ordered 3-candidate-per-role preference list plus a collision guard enforced again at the filter-builder level (not just trusted from the config-builder), and the config-assembly logic was extracted into a pure, snapshot-tested function rather than left inline.
- **ACP opus/sonnet tier gap**: the first `interactive_claude_code` slug map only distinguished `subagent_model`/`haiku_model`; review flagged that `opus_model`/`sonnet_model` are configured *tiers of main-thread traffic* (Claude Code's own role swap), not delegation — a distinct opus/sonnet slug was falling through to `unknown` instead of `root`. Fixed by adding both canonical names to `root_slugs`, with a test exercising a distinct opus/sonnet config.
- **docs accuracy fixes**: a review pass on `docs/sub-agent-context.qmd` caught overclaiming opencode/ACP attribution guarantees relative to what was actually live-verified (the table above is the corrected version — e.g. opencode's custom-Task-tool-agent gap and the "best-effort, not independently live-verified" language for title/summary/compaction).

### Review round 2 (2026-09-03)

Addressed all 9 inline findings and 3 of 4 non-blocking cleanups from the 2026-08-19 review (see thread replies; commits f2ca2c3..c6248ed), and merged current `main` (6a574dc). Notable outcomes beyond the literal findings: the anthropic OpenCode collision alternates were Claude 3.x ids no longer in the models.dev catalog (now single-candidate); `build_opencode_model_aliases` was a regression, not a no-op (`get_model()` rejected bare role names); `interactive_claude_code` now presents a distinct subagent slug by default.

- Fixes authored by five Claude Fable 5.1 subagents (fresh context each, one per area), each required to reproduce the finding with a failing test before changing code; one finding was narrowed with evidence (V2 codex FINAL_ANSWER already classified `root`; only the span close moved to filter time).
- Verification on the merged branch: fast suite 440 passed / 48 skipped / 1 failed (`test_bundled_catalog_tracks_live_latest`, a live codex catalog drift test that failed identically at baseline before these changes); `ruff`, `ruff format --check`, and `mypy --strict` clean against an inspect_ai build of the upstream `feature/agent-bridge-context` branch. Both live drift tests (kimi, gemini) pass locally today.

### Review round 3 (2026-09-04, clean context)

Two further fresh-context passes over the whole branch after round 2: a general reviewer with no prior context, then `/code-review` at high effort (Claude Fable 5.1 in both cases). 11 findings — 10 fixed in `1a974b7` and `30827ef` (details in the PR comment of the same date), 1 deferred (codex `classify`/`on_pending` double orchestration, an altitude cleanup), plus one design decision left open for maintainers (OpenCode sentinel pinned to a catalog id with no opt-out). The two most consequential fixes: `LiveConsumer.classify` now checks the main-thread slugs before the pending-prompt heuristic (a verbatim delegated prompt previously stamped the root thread `subagent`), and gemini primary aliases (`flash`, `pro`, `auto`, ...) resolve to root slugs (previously the root thread classified `utility`). Final tree: 467 passed / 48 skipped / 1 pre-existing live-catalog failure; ruff and strict mypy clean.


src/inspect_swe/acp/_agents/claude_code/claude_code.py
```diff
@@ -22,6 +28,47 @@
...
+    if opus_model is not None:
+        root_slugs.add(get_model(opus_model).canonical_name())
+    if sonnet_model is not None:
+        root_slugs.add(get_model(sonnet_model).canonical_name())
+
+    kind_by_slug: dict[str, Literal["subagent", "utility"]] = {}
+    if subagent_model is not None:
+        kind_by_slug[get_model(subagent_model).canonical_name()] = "subagent"
+    if haiku_model is not None:
+        kind_by_slug[get_model(haiku_model).canonical_name()] = "utility"
```
craig-meridian: [P2] This builder has no counterpart to the distinctness invariant `resolve_claude_code_models` enforces, and both collision directions misclassify silently:

- `subagent_model == haiku_model` (the natural "one cheap model for background AND subagents" config the native path explicitly handles in `model.py`): the two inserts share a canonical name, so this line overwrites the `"subagent"` entry — every genuine Task-subagent request stamps `"utility"` and `is_sub_agent()` returns `False`.
- `subagent_model` resolving to the same slug as `default_model`: the slug lands in both `root_slugs` and `kind_by_slug`, root wins (`slug_map_classifier` checks it first), so subagent traffic stamps `"root"` and a filter gated on `is_root_agent()` actively steers sub-agents — the harm this PR exists to prevent.

Since `_start_agent` exports these same names as `CLAUDE_CODE_SUBAGENT_MODEL`/`ANTHROPIC_SMALL_FAST_MODEL`, colliding config makes the traffic genuinely wire-indistinguishable — so the fix available here is the native path's synthetic-name technique (present a distinct subagent alias), or at minimum an honest ambiguous result plus a logged warning like `build_opencode_filter`'s. Tests for the subagent/haiku and subagent/root collisions would pin whichever behavior you pick.

src/inspect_swe/_claude_code/_events/live_consumer.py
```diff
@@ -239,29 +286,135 @@ def process_jsonl_line(self, raw: dict[str, Any]) -> None:
...
+        slug = request.model if request is not None else None
+        self._check_subagent_slug_drift(slug)
+
+        if slug == self._models.subagent:
+            return AgentBridgeContext("subagent")
+        if slug == self._models.haiku and self._models.haiku != self._models.presented:
+            return AgentBridgeContext("utility")
+        if self._match_pending_prompt(messages) is not None:
+            return AgentBridgeContext("subagent")
+        if slug == self._models.presented:
```
craig-meridian: [P2] The truth table never consults `self._models.opus` / `self._models.sonnet`, but `env.py` exports `ANTHROPIC_DEFAULT_OPUS_MODEL`/`ANTHROPIC_DEFAULT_SONNET_MODEL` with those slugs, so CC main-thread traffic can legitimately arrive under them (e.g. an opusplan tier swap — which `claude_code`'s own docstring supports). With a distinct `opus_model` configured and a Task sub-agent pending, such a request matches neither branch 1–4 and falls to branch 5 → `"unknown"` instead of `"root"`. A consumer following the docs' own advice (check `kind == "root"` explicitly for positive confirmation) then skips steering genuine main-thread turns — and each such request also advances the drift-canary counter.

The ACP sibling (`build_claude_code_acp_filter`) already adds both canonical names to `root_slugs` for exactly this reason; branch 4 here should do the same (`slug in {presented, opus, sonnet}` modulo the inherit-presented cases).

src/inspect_swe/_opencode/opencode.py
```diff
@@ -33,6 +39,397 @@
...
+            "claude-3-5-haiku-20241022",
+            "claude-3-haiku-20240307",
+        ),
+        (
+            "claude-3-5-haiku-20241022",
+            "claude-3-haiku-20240307",
+            "claude-3-opus-20240229",
+        ),
+    ),
+    "openai": (
```
craig-meridian: [P2] The openai/google sentinels are injected as `agent.general/explore/scout.model` with no runtime validation or fallback, and caveat 3's own live evidence shows what a catalog-rejected id does: the sub-agent's Task call fails outright with `"Model not found"`. So for `opencode_model="openai/..."` or `"google/..."` on any OpenCode version whose catalog lacks the chosen sentinel (these lists are self-described UNVERIFIED — and the same applies to anthropic on an older pinned `version=`), built-in delegation that worked before this PR hard-fails. Worse, it's undetectable: a rejected model never produces a bridged request, so no filter, classifier, or log line can see it — a classification-only feature introducing a silent behavioral regression.

Suggest validating sentinel resolvability against the *installed* OpenCode at setup (skip injection when a sentinel isn't resolvable, like the unrecognized-provider path already does), and/or a "Task usage observed but no sentinel slug ever seen" canary symmetric with claude_code's.

src/inspect_swe/_codex_cli/_events/consumer.py
```diff
@@ -192,6 +196,47 @@ def on_complete(self, event: ModelEvent) -> None:
...
+        sub-agents is `_attribute`'s own ambiguity fallback, not a genuine
+        root signal) — otherwise "unknown".
+        """
+        request = current_bridge_request()
+        if request is not None and request.model == GUARDIAN_MODEL_SLUG:
+            return AgentBridgeContext("utility")
+        if is_compaction_request(messages):
+            return AgentBridgeContext("utility")
+
+        self._harvest_bindings(messages)
```
craig-meridian: [P2] `classify` harvests bindings but never processes the thread-close signals (`completed_thread_ids` / `final_answer_authors`) that `on_pending` handles, so it attributes against a stale open-agent set. V1 flow: the last open sub-agent completes, and the root thread's next request carries the completion notification in its own input. `classify` runs first (per this PR's own sequencing test), sees `self._agents` still non-empty, gets no recipient match (the `{"/root"}` rescue only fires for V2 `agent_message` items on `ContentText.internal` — V1 completion arrives as `wait_agent`/`close_agent` results or `<subagent_notification>` text, which produce an empty recipient set), and stamps `"unknown"` for a genuinely-root request whose input contains the very signal proving the set should be empty. One misattributed root turn per V1 sub-agent completion — consistent with the 4 `unknown` root calls in validation run b.

Processing close signals in `classify` before attributing (as `on_pending` does) would convert those to honest `"root"`.

src/inspect_swe/_claude_code/model.py
```diff
@@ -74,7 +90,54 @@ def role_name(role_model: str | None) -> str:
...
+    # already be resolved above so all four names are known here.
+    if subagent_model is None:
+        subagent_name = presented
+        subagent_route: Model = served_model
+    else:
+        subagent_route = get_model(subagent_model)
+        subagent_name = subagent_route.name
+
+    if subagent_name in {presented, opus, sonnet, haiku}:
+        subagent = f"{presented}-subagent"
```
craig-meridian: [P2] The synthetic name isn't re-checked against `{presented, opus, sonnet, haiku}`, so the invariant the docstring promises can still be violated. Constructible (reproduced with mockllm): `resolve_claude_code_models("mockllm/model", None, opus_model="mockllm/model-subagent")` → `opus == subagent == "model-subagent"`, and this branch's `aliases[subagent] = subagent_route` overwrites the caller's explicit opus alias with the primary served model — main-thread opus requests are then routed to the wrong model *and* classified `"subagent"` (the subagent slug is checked first in `classify`). Admittedly a contrived config — no real catalog name has the `<presented>-subagent` shape — but a re-check of the synthetic name against the full role set (extending the suffix on collision) closes it in a couple of lines, plus a per-role collision test.

src/inspect_swe/_opencode/opencode.py
```diff
@@ -33,6 +39,397 @@
...
+    ),
+}
+
+OPENCODE_BUILTIN_SUBAGENTS: tuple[str, ...] = ("general", "explore", "scout")
+"""Built-in OpenCode subagents (opencode.ai/docs/agents/), routed to the
+subagent sentinel so their traffic is slug-distinguishable from the main
+thread even when invoked without a `model:` override of their own.
+Live-verified (2026-08-08, `general`): the per-agent override IS honored."""
+
+OPENCODE_UTILITY_AGENTS: tuple[str, ...] = ("title", "summary", "compaction")
```
craig-meridian: [P3] Caveat 2 documents the spawnable-agent hazard for `general`/`explore`/`scout`, but it applies to these three names too — and here it's worse than under-attribution: on an install where `title`/`summary`/`compaction` aren't reserved `AgentConfig` names, each entry creates a real, prompt-less, mode-`"all"` agent, and a main agent that spawns "summary" gets a genuine delegated sub-agent whose requests carry the small-model sentinel and stamp `"utility"` — so `is_sub_agent()` misses real delegation. Current config.json does name all three as agent keys, so the default `version="auto"` path looks safe, but the injection is acknowledged UNVERIFIED and the schema-recognizes vs. runtime-hides gap is untested. Worth either spending the follow-up live run before relying on it, or gating the utility trio off until verified (keeping the verified subagent overrides).

src/inspect_swe/_kimi_code/kimi_code.py
```diff
@@ -103,6 +85,78 @@ def _is_legacy_str_filter(fn: GenerateFilter) -> bool:
...
+    the marker — tool output echoing kimi's own docs/source therefore can't
+    flip a main-thread request to utility (the harmful failure direction; a
+    missed match merely reverts to root, which no gate acts on differently).
+    Substring rather than prefix match: the rendered instruction opens with a
+    preamble sentence ("You are about to run out of context. ...") before the
+    marker phrase.
+    """
+    if (
+        messages
+        and isinstance(messages[-1], ChatMessageUser)
```
craig-meridian: [P3] The guard rules out tool-output echoes but not a *genuine* user message containing the phrase: an eval whose task prompt (or centaur interjection) legitimately says "Write a first-person handoff note for the next engineer…" classifies `"utility"` on every turn where that message is last, so `is_root_agent()` gates skip exactly the turns they should govern — the harmful direction the docstring says is prevented. Narrow trigger, but concrete. Matching more of the rendered instruction (the "You are about to run out of context." preamble plus the marker) or requiring the message to *begin* with the preamble would keep the drift-tolerance while making task-prompt collisions implausible.

src/inspect_swe/_claude_code/_events/live_consumer.py
```diff
@@ -239,29 +286,135 @@ def process_jsonl_line(self, raw: dict[str, Any]) -> None:
...
+            self._subagent_slug_seen = True
+            return
+        if (
+            not self._first_subagent_span_seen
+            or self._subagent_slug_seen
+            or self._subagent_slug_drift_warned
+        ):
+            return
+
+        self._requests_since_first_subagent_span += 1
```
craig-meridian: [P3] Spurious-warning path: `on_complete` sets `_first_subagent_span_seen` the moment a Task tool_call appears in model *output* — before any sub-agent request exists. If the Task dies inside Claude Code (bad `subagent_type`, immediate tool error) and no sub-agent request ever reaches the bridge, the next 5 ordinary main-thread `classify` calls increment this counter and emit the "CC version drift or slug rejection likely" warning — asserting subagent traffic never carried the slug when there was no subagent traffic at all. Log-only, but it points operators at a nonexistent CC drift problem. Gating the increment on `self._pending_subagents` being non-empty would confine the canary to windows where subagent traffic is actually expected.

src/inspect_swe/_gemini_cli/models.py
```diff
@@ -0,0 +1,69 @@
...
+subagent definition's `modelConfig.model` (see `agents/local-executor.js`)
+defaults to the *same* `DEFAULT_GEMINI_MODEL` used as gemini-cli's own
+primary model when unset, and to whatever else a subagent definition names
+it when set. There is no known-name to key a `"subagent"` classification on,
+so this module doesn't attempt one.
+"""
+
+from typing import Literal
+
+GEMINI_UTILITY_MODEL_SLUGS: frozenset[str] = frozenset(
```
craig-meridian, line 75: [P3] This slug set is pinned to `gemini-cli-core@0.54.4`'s alias table while `version="auto"` installs whatever is latest, and unlike its siblings it got no drift protection at all — claude_code got a runtime canary, kimi got a live drift test; gemini got "re-verify if suspected stale", but nothing ever suspects for you. When a routine gemini-cli release retargets an alias, utility calls classify `"unknown"` — and `is_root_agent()` deliberately treats unknown as root, so steering/early-stop filters silently resume firing on loop-detection/next-speaker/compaction calls. Deeper fix available at setup time: the installed package ships `defaultModelConfigs.js`, so the slug set could be read per-install (structural, drift-immune). At minimum, a kimi-style drift test plus a "never saw any utility slug" canary.

craig-meridian (commented):
Nice, I can see the value of this! Below is a review synthesised from reviews by Fable 5 and GPT 5.6 Sol. Feel free to push back on anything.

---

Impressive piece of work — the per-agent structural signals are well chosen, the honest-`unknown` policy is the right call, and the live-validation evidence made this reviewable. I found a handful of classification-correctness edges, mostly places where one path enforces an invariant its sibling path doesn't. Five P2s and four P3s inline.

Non-blocking cleanup observations, take or leave:
- `build_gemini_acp_filter` is a body-identical re-implementation of `build_gemini_filter` — it could just delegate, so future changes to gemini classification can't drift between the two variants.
- `CodexConsumer.classify` re-runs `_harvest_bindings` / `agent_message_recipients` / `is_compaction_request` / `_attribute` that `on_pending` then repeats for the same request — a growing double scan per request on long conversations; a per-request memo (or an early-out when `self._agents` and `self._thread_index` are empty) would remove it.
- The root-vs-sentinel collision guard in `build_opencode_filter` re-states precedence `slug_map_classifier` already guarantees; if the guard+warning lived in `slug_map_classifier` itself, every caller (including the ACP claude_code path flagged below) would get it for free.
- `build_opencode_model_aliases` is a no-op by its own docstring (the bridge fallback already serves unaliased slugs) — a comment would do the same job with less machinery.

src/inspect_swe/acp/_agents/claude_code/claude_code.py
dragonstyle: Fixed in 29fdb03. `resolve_claude_code_acp_models` now applies the native path's invariant: the subagent role is presented as `<presented>-subagent` (next free `-N` suffix if a role already owns that name) and aliased to the intended served model, and `_start_agent` exports that slug as `CLAUDE_CODE_SUBAGENT_MODEL`. Both collision directions are pinned by tests (subagent==haiku, subagent==default, plus haiku==default staying root with no warning). Not shared with `resolve_claude_code_models` because the ACP path keys the bridge on `canonical_name()` and takes `Model` instances whose bound config must survive; the docstring records that. Side effect worth noting: with no `subagent_model` the ACP variant now presents a distinct subagent slug by default, same as native, so its Task sub-agent traffic is attributable out of the box (docs row updated).

src/inspect_swe/_claude_code/_events/live_consumer.py
dragonstyle: Fixed in f2ca2c3. Branch 4 is now `slug in {presented, opus, sonnet}`; the truth-table docstring says why (tiers of the main thread, not delegation). Reproduced first: distinct `opus_model`, one pending Task, request slug == opus gave `unknown`. Tests added for the opus and sonnet cases.

src/inspect_swe/_opencode/opencode.py
dragonstyle: Fixed in 93d4502, and the finding is wider than stated. Only the anthropic sentinel is injected now; the openai/google candidates are kept in `_UNVERIFIED_SENTINEL_CANDIDATES` but never consulted, so those providers take the same warn-once, no-injection path as an unrecognized provider. While checking, the anthropic collision alternates (`claude-3-5-haiku-20241022`, `claude-3-haiku-20240307`) turned out to be absent from the current models.dev catalog, so they would have hard-failed the same way when a caller's `opencode_model` collided with the default sentinel; the subagent role is now single-candidate and a collision omits the override. Runtime validation against the installed OpenCode's catalog is the follow-up that would let openai/google be re-enabled; not attempted here.

src/inspect_swe/_codex_cli/_events/consumer.py
dragonstyle: Fixed in 8d1bb5d. Reproduced with a V1 flow (spawn, bind, then `wait_agent` completion in the root's next request): `classify` returned `unknown`. `on_pending` and `classify` now share `_sync_threads` (harvest bindings, close completed and FINAL_ANSWER threads), so the request classifies `root`. Closing pops the thread, so the later `on_pending` pass emits no second `SpanEndEvent`; the existing sequencing test still passes and the new tests assert single emission. Under V2 the FINAL_ANSWER `agent_message` already hit the `/root` rescue, so the only V2 change is that the child span now closes at filter time.

src/inspect_swe/_claude_code/model.py
dragonstyle: Fixed in f2ca2c3. Reproduced your mockllm case (opus alias silently overwritten with the primary). The synthetic name is now re-checked against all four role names and takes the next free suffix (`-subagent`, `-subagent-2`, ...); since the alias keys at that point are exactly the role names, no caller alias can be overwritten. Parametrized per-role collision tests added, plus one where two suffixes are already taken.

src/inspect_swe/_opencode/opencode.py
dragonstyle: Fixed in 93d4502. The utility trio is no longer injected as `agent.<name>.model` overrides; the documented `small_model` setting stays. `OPENCODE_UTILITY_AGENTS` is kept with a comment on what live verification would re-enable, and the config snapshot tests were updated.

src/inspect_swe/_kimi_code/kimi_code.py
dragonstyle: Fixed in 0fa2c8e. The classifier now requires the last user message to open with the instruction's first sentence ("You are about to run out of context.") and contain the marker, whitespace-normalized. Verified against upstream main that the rendered instruction has nothing prepended, so the prefix match is sound. Tests: genuine instruction is utility; the marker phrase mid-prompt without the preamble is root; the full instruction inside a tool message is root. The live drift test now checks the preamble too.

src/inspect_swe/_claude_code/_events/live_consumer.py
dragonstyle: Fixed in f2ca2c3. Counting is gated on `_pending_subagents` being non-empty, the window in which sub-agent traffic is actually expected; `_handle_user` closes it on the Task's tool_result. Reproduced first: Task in output, tool_result with no sub-agent request, ten root requests produced one warning. `_first_subagent_span_seen` became redundant and was removed; the counter is renamed to match what it measures.

src/inspect_swe/_gemini_cli/models.py
dragonstyle: Addressed in 0fa2c8e with a kimi-style live drift test rather than per-install reading. `tests/test_gemini_models_drift.py` fetches gemini-cli's `defaultModelConfigs.ts` from upstream main, resolves each utility alias through the `model`/`extends`/routing-table chain, and fails naming the alias and slug if any resolves outside `GEMINI_UTILITY_MODEL_SLUGS` (github-action-gated, skips offline, same as the kimi test). Re-verified today against gemini-cli-core 0.58.0: all fifteen utility aliases still resolve to the three slugs in `models.py`, and the provenance note is updated. Reading the installed package's alias table at setup is a possible later improvement.

dragonstyle:
On the four non-blocking observations:

- `build_gemini_acp_filter`: deleted; the ACP variant now calls `build_gemini_filter` directly (29fdb03).
- `CodexConsumer.classify` double scan: `classify` now early-outs to `root` when no sub-agents are open, which removes the scan for the common case; the harvest/close pass is shared with `on_pending` via `_sync_threads` and is idempotent (8d1bb5d). No cross-call memo keyed on message identity.
- Root-vs-sentinel collision guard: moved into `slug_map_classifier`, which drops colliding `kind_by_slug` entries with one warning at construction, so every caller gets it (29fdb03). `build_opencode_filter`'s own guard is now redundant and can go in a follow-up; left in place to keep the opencode change focused on the P2/P3 items.
- `build_opencode_model_aliases`: removed (93d4502). It was not just a no-op: the `get_model(model)` it required rejected bare role names such as `model="grader"`, which the bridge fallback resolves.

All 9 inline findings are addressed in commits f2ca2c3 through c6248ed; each thread has a reply with what changed. Branch is also merged up to current `main` (6a574dc).

dragonstyle:
## Clean-context review round 3

Two independent fresh-context passes (a reviewer with no prior context, then the repo's `/code-review` at high effort) over the whole branch after round 2. Findings and outcomes, in commits `1a974b7` and `30827ef`:

**Fixed (P2)**
- `LiveConsumer.classify` ran the pending-prompt substring match before the presented/opus/sonnet slug check. A root request whose task prompt contains a delegated Task prompt verbatim (common) was stamped `subagent` for the whole delegation window, the exact failure an `is_root_agent()` gate exists to prevent. Slug checks now precede the heuristic, and `_attribute` applies the same rule so span attribution agrees. (`1a974b7`)
- `effort` was not applied to an explicit `subagent_model` (regression vs main). (`1a974b7`)
- `interactive_codex_cli` docstring and docs claimed sub-agent traffic falls through to `unknown`; it classifies `root` because `spawn_agent` has no per-agent model. Now stated plainly. (`1a974b7`)
- `gemini_model="gemini-3-pro-preview"` (or any other presented model that is also a utility slug) tripped the collision warning on every sample. Dropped from the kind map as documented under-attribution. (`1a974b7`)
- gemini-cli resolves primary aliases (`auto`/`pro`/`flash`/`flash-lite`/...) to concrete slugs before the wire, several of which are utility slugs, so `gemini_model="flash"` classified the root thread `utility` for the whole episode. Each alias's possible targets (from gemini-cli-core 0.58.0's routing table, now also checked by the live drift test) are root slugs. (`30827ef`)
- With `subagent_model` unset, a caller `model_aliases` override of the presented name rerouted main-thread traffic but left Task sub-agent traffic on the un-overridden model (on main both shared one slug). The synthetic slug now follows the override; same rule in the ACP resolver. (`30827ef`)

**Fixed (P3 / cleanup)**
- Drift canary counted every non-subagent request while a Task was pending, including retries, so ordinary runs could warn falsely. It now counts only prompt-matched sub-agent requests missing the slug. (`30827ef`)
- Native small-fast check now matches the ACP variant: haiku sharing any main-thread slug is root. (`1a974b7`)
- The `<presented>-subagent[-N]` collision loop is one helper, `distinct_subagent_name`, used by both resolvers. (`30827ef`)
- `classify_filter` re-emits the legacy str-first filter deprecation the bridge can no longer observe. (`30827ef`)
- Removed `build_opencode_filter`'s duplicate collision guard and the inert `_UNVERIFIED_SENTINEL_CANDIDATES` / `OPENCODE_UTILITY_AGENTS` constants. (`30827ef`)

**Not changed, needs a decision**
- OpenCode built-in sub-agent routing is pinned to the catalog id `claude-haiku-4-5-20251001` with no opt-out. OpenCode refreshes its catalog from models.dev at runtime, so if that id is retired, `general`/`explore`/`scout` delegation fails silently under a fixed inspect_swe version. Options: an opt-out parameter on `opencode()`, validating against `opencode models` in the sandbox at setup, or a synthetic model declared under `provider.<id>.models` (would need a live probe; P2's earlier probe did not declare the model there).
- Codex: `classify` and `on_pending` still run two orchestrations over the same request (idempotent, sub-millisecond); a single memoized observe step would collapse the `/root`-recipient rule into one place. Deferred as an altitude cleanup.
- The import-floor blocker is unchanged: the branch needs an inspect_ai release containing UKGovernmentBEIS/inspect_ai#4791 (still open); CI is red on that alone.

Verification on the final tree: 467 passed, 48 skipped, 1 failed (`test_bundled_catalog_tracks_live_latest`, live codex catalog drift, failing identically at baseline); ruff, ruff format, and strict mypy clean against an inspect_ai build of upstream's `feature/agent-bridge-context`. Every fix has a test that fails on the pre-fix source.

src/inspect_swe/_codex_cli/_events/consumer.py
```diff
@@ -192,10 +191,73 @@ def on_complete(self, event: ModelEvent) -> None:
...
+        root signal) — otherwise "unknown".
+        """
+        request = current_bridge_request()
+        if request is not None and request.model == GUARDIAN_MODEL_SLUG:
+            return AgentBridgeContext("utility")
+        if is_compaction_request(messages):
+            return AgentBridgeContext("utility")
+
+        # nothing open → nothing to bind, close, or attribute against
+        if not self._agents:
```
craig-meridian, line 232: [P2] Preserve V2 sub-agent identity across follow-up turns

After a V2 child sends `FINAL_ANSWER`, `_sync_threads` removes its identity. If the parent then calls `followup_task`, the child's next request classifies as `root`, even though its `agent_message` recipient still identifies `/root/worker`. I reproduced this sequence. Consequently, `is_root_agent()` allows root steering and early-stop logic into the resumed child. Preserve identity independently of open spans, or classify the explicit non-root recipient before taking this shortcut.

(With another sub-agent still open the same request lands on `unknown` instead, which is only marginally better.)

Suggested fix: treat any `agent_message` recipient other than `/root` as a positive sub-agent signal, checked before the `if not self._agents` shortcut and before span state. A request can only carry an inbound message addressed to a non-root path if it came from that sub-agent, so this needs no thread bookkeeping to survive a close. It also covers a V2 child whose spawn the sink never saw (see the `codex_cli.py` comment on filter-returned output). One existing expectation changes: `test_classify_ambiguous_falls_to_unknown` would report `subagent` for two same-named spawns, which is correct — the kind is knowable even when the span identity isn't.

src/inspect_swe/_codex_cli/_events/consumer.py
```diff
@@ -192,10 +191,73 @@ def on_complete(self, event: ModelEvent) -> None:
...
+            return AgentBridgeContext("root")
+
+        self._sync_threads(messages)
+        if agent_message_recipients(messages) == {"/root"}:
+            return AgentBridgeContext("root")
+
+        span_id = self._attribute(messages)
```
craig-meridian, line 239: [P2] Do not treat a V1 prompt substring as conclusive sub-agent identity

The root-prompt collision fixed for Claude Code in review round 3 remains in Codex's V1 path. When the root delegates a sentence copied from its task prompt, `_attribute` matches that sentence in the root's subsequent request and `classify` returns `subagent`. I reproduced this with the root's original prompt, spawn call, and spawn result. Root steering then stops throughout the delegation window. Track a positive parent signal, or return `unknown` when the substring also belongs to the parent history.

Suggested fix: the parent signal is already in the request. When substring-matching, drop any candidate whose spawn `call_id` appears in the request's own history (as an assistant `spawn_agent` tool call or its `ChatMessageTool` result) — the requester is the spawner, never the spawned agent, and a V1 child's thread never contains its own spawn call. With the only candidate excluded, root classifies `unknown` for the window (the documented honest fallback) rather than `subagent`. Please also apply this in `_attribute` itself so span attribution agrees, as 1a974b7 did for `LiveConsumer`.

More generally: this and the OpenCode alias comment are both sibling-agent analogues of bugs round 3 fixed in Claude Code only. Before the next round is called done, could you sweep each classifier fix across the other agents' equivalents?

src/inspect_swe/_opencode/opencode.py
```diff
@@ -121,17 +432,31 @@ def opencode(
...
         port = store().get(MODEL_PORT, 3000) + 1
         store().set(MODEL_PORT, port)
 
+        bridge_model = f"inspect/{model}" if model is not None else "inspect"
+
         async with sandbox_agent_bridge(
             state,
-            model=model,
+            model=bridge_model,
+            # the sentinel slugs need no alias entries: the bridge's fallback
```
craig-meridian, line 455: [P2] Route injected OpenCode sentinels through the effective primary alias

The fallback does not preserve routing when `model_aliases` overrides the primary slug. For example, with `model="mockllm/fallback"` and `model_aliases={"claude-sonnet-4-5": "mockllm/primary-override"}`, root requests use `primary-override`, but the injected Haiku sub-agent slug resolves to `fallback`. Previously, default sub-agents inherited the primary slug and its alias, so this classification-only feature silently changes the model used for delegated work. Make the sentinel follow the primary's effective route while preserving explicit caller overrides.

Suggested fix: mirror what `resolve_claude_code_models` does since 30827ef (`aliases[subagent] = aliases[presented]`). In `execute`, when the caller's `model_aliases` contains the bare primary slug and does not itself map the sentinel, add the sentinel's bare id with the same target; pass the augmented dict to `sandbox_agent_bridge`. Both the subagent and small-model sentinels need this. The comment above the bridge call ("the sentinel slugs need no alias entries") and the module docstring's matching paragraph should be revised, since they describe the un-aliased case only. A test with a caller alias on the primary slug that asserts the sentinel resolves to the same target would pin it.

src/inspect_swe/_codex_cli/codex_cli.py
```diff
@@ -377,7 +378,7 @@ async def execute(state: AgentState) -> AgentState:
                     default=get_model(model) if transparent_proxy else None,
                 ),
                 forward_generation_config=transparent_proxy,
-                filter=filter,
+                filter=classify_filter(filter, consumer.classify),
```
craig-meridian, line 381: [P2] Register Codex spawns when a filter returns model output

A supported filter that calls `model.generate()` itself and returns `ModelOutput` bypasses the bridge's `ModelEventSink`, which is active only around the bridge's default generation. `CodexConsumer.on_complete()` therefore never registers a `spawn_agent` output, and subsequent V1 child requests classify as `root`, allowing root-only steering or termination to affect sub-agents. Reproduced with real `bridge_generate` and MockLLM: normal generation produced `['root', 'subagent']`; filter-driven generation produced `['root', 'root']` with zero tracked agents. Ensure returned outputs update delegation tracking before the response reaches Codex.

Suggested fix, if handled here: give `classify_filter` an optional callback invoked with `(messages, output)` when the user filter returns a `ModelOutput`, and have `CodexConsumer` expose a method that runs the spawn-registration and `close_agent` block from `on_complete` against that output. That keeps the change inside inspect_swe. Claude Code degrades gracefully in this scenario (the subagent slug still classifies; only the prompt-match fallback goes missing), so Codex is the only agent that loses classification outright.

If deferred: the recipient-first rule suggested on `consumer.py` already covers the V2 case, so the documented gap would be V1 only. Either way please note it in the `codex_cli` row of the guarantees table.

craig-meridian (commented):
Thanks Charles! Review below. I appreciate we don't want the review/fix cycle going on forever, so if any of these seem unreasonable, please do push back, but I think we're close to converging on just P3s or nits.

---

Four P2s, all Codex/OpenCode. Two of them (the V1 prompt collision and the OpenCode sentinel alias) are the sibling-agent analogues of bugs round 3 fixed for Claude Code only.

Disposition: the V1 prompt collision fires in the harmful direction (root steering goes dark for the delegation window) and the OpenCode alias routing is a regression against main, so both are needed before merge; the V2 follow-up identity is a small change once you're in the file. The filter-returned-`ModelOutput` case should be fixed now if the change stays inside inspect_swe (a sketch is in the comment); if the proper fix belongs in inspect_ai, documenting the limitation in the guarantees table and tracking it separately is sufficient.

Each comment carries a suggested fix. All four were re-verified against 30827ef on an inspect_ai build of upstream #4791 (fast suite, strict mypy and ruff otherwise clean there).

PR #119, opened 2026-08-11
feat: claude_code reasoning effort option and bridged-MCP allowlist opt-out

sjawhar:
Two small claude_code agent options, combined per our fork maintainer's request:

**Reasoning effort**: `claude_code()` gains an `effort` option (`"low"`/`"medium"`/`"high"`/`"xhigh"`/`"max"`) that sets `reasoning_effort` on the `GenerateConfig` of every served model the agent resolves, so evals can control the served model's thinking budget per agent. This is applied host-side rather than passed through as a Claude Code CLI flag: the bridge drops the inner agent's own request-level generation config by default, so a CLI-level flag would have no effect on the model actually serving the request. Covered by `tests/test_claude_code_effort.py`.

**Bridged-MCP allowlist opt-out**: `permission_mode="auto"` callers can suppress the bridged MCP allowlist (`--allowed-tools` restriction) that claude_code applies to bridged tools by default, via `allowlist_bridged_tools=False`. The default stays unchanged; the opt-out is for evaluations that specifically measure Claude Code's own auto-mode classifier acting on bridged tools. Construction raises `ValueError` if `allowlist_bridged_tools=False` is combined with any `permission_mode` other than `"auto"`, since every other mode denies an excluded bridged tool without prompting rather than adjudicating it. Covered by `tests/test_claude_code_mcp_allowed_tools.py`.

`make check` clean (ruff, mypy); targeted suites pass.

Agent involvement disclosure: branch assembled and PR opened by Claude on Sami Jawhar's behalf; reworked in response to review feedback from epatey (the effort mechanism moved from a CLI passthrough to a host-side GenerateConfig override after epatey verified the CLI flag had no effect on the served model, and the allowlist option gained a construction-time guard).

### DIFF
--- a/src/inspect_swe/_claude_code/claude_code.py
+++ b/src/inspect_swe/_claude_code/claude_code.py
@@ -64,6 +64,7 @@ from .model import resolve_claude_code_models
 ClaudeCodePermissionMode = Literal[
     "acceptEdits", "auto", "bypassPermissions", "default", "dontAsk", "plan"
 ]
+ClaudeCodeEffort = Literal["low", "medium", "high", "xhigh", "max"]
 
 
 class ClaudeCodeDeprecatedArgs(TypedDict, total=False):
@@ -128,6 +129,7 @@ def claude_code(
     attempts: int | AgentAttempts = 1,
     model: str | None = None,
     model_config: str | None = None,
+    effort: ClaudeCodeEffort | None = None,
     model_aliases: dict[str, str | Model] | None = None,
     opus_model: str | None = None,
     sonnet_model: str | None = None,
@@ -185,6 +187,7 @@ def claude_code(
             bridged to the served Inspect model regardless. (Claude Code renders
             the genuine name/cutoff for recognized Anthropic ids and shows other
             ids verbatim.)
+        effort: Claude Code reasoning effort. ``None`` leaves the CLI default.
         model_aliases: Optional mapping of model names to Model instances or model name strings.
             Allows using custom Model implementations (e.g., wrapped Agents) instead of standard models.
             When a model name in the mapping is referenced, the corresponding Model/string is used.
@@ -317,11 +320,7 @@ def claude_code(
                 if effective_permission_mode is not None
                 else ["--dangerously-skip-permissions"]
             )
-            cmd = [
-                *permission_flag,
-                "--model",
-                models.presented,
-            ]
+            cmd = claude_code_command(permission_flag, models.presented, effort)
 
             # add interactive options if not running as centaur
             if centaur is False:
@@ -615,6 +614,18 @@ def _system_prompt_args(
         args.extend(["--append-system-prompt", "\n\n".join(system_texts)])
 
     return args
+
+
+def claude_code_effort_args(effort: ClaudeCodeEffort | None) -> list[str]:
+    if effort is None:
+        return []
+    return ["--effort", effort]
+
+
+def claude_code_command(
+    permission_flag: list[str], model: str, effort: ClaudeCodeEffort | None
+) -> list[str]:
+    return [*permission_flag, "--model", model, *claude_code_effort_args(effort)]
 
 
 async def _seed_claude_config(
--- a/src/inspect_swe/_claude_code/claude_code.py
+++ b/src/inspect_swe/_claude_code/claude_code.py
@@ -147,6 +147,7 @@ def claude_code(
     debug: bool | None = None,
     replace_system_prompt: str | None = None,
     allowlist_mcp_tools: bool = True,
+    allowlist_bridged_tools: bool = True,
     **deprecated_args: Unpack[ClaudeCodeDeprecatedArgs],
 ) -> Agent:
     """Claude Code agent.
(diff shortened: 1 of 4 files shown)


src/inspect_swe/_claude_code/claude_code.py
```diff
@@ -145,6 +147,7 @@ def claude_code(
     debug: bool | None = None,
     replace_system_prompt: str | None = None,
     allowlist_mcp_tools: bool = True,
+    allowlist_bridged_tools: bool = True,
```
epatey, line 149: Please raise a `ValueError` at construction when `allowlist_bridged_tools=False` is combined with any `permission_mode` other than `"auto"`. In every other mode the result is bridged tools silently denied without prompting, so the eval completes toolless and gets scored. The docstring already restricts this to auto mode; failing fast matches the existing `auto_mode`/`permission_mode` conflict check in `resolve_claude_code_deprecated_args`.

epatey (changes requested):
Requesting changes, and asking to split this back into two PRs. The two halves are in very different states.

**`allowlist_bridged_tools`: close to mergeable.** The motivation is solid and the default preserves existing behavior. One ask (inline): raise a `ValueError` at construction when `allowlist_bridged_tools=False` is combined with any `permission_mode` other than `"auto"`. In every other mode the result is bridged tools silently denied and the eval scored toolless. The docstring already says "only with auto"; the guard makes it fail fast, matching the existing `auto_mode`/`permission_mode` conflict check. Split this out and it can land quickly.

**`effort`: needs rethinking before it's mergeable.** I ran this end-to-end from the PR head (Claude Code 2.1.231, Docker sandbox, filter capturing the bridged request). The flag only changes the CLI's request under adaptive-thinking model identities (opus-4-6 yes, sonnet-4-5 no), and in every case the bridge then drops it: `claude_code` never sets `forward_generation_config`, so `clear_generation_params` strips `effort`/`reasoning_tokens` before generation. As written the option has no effect on the served model in any reachable configuration. The bridge treats generation tuning as the Inspect model's province by design, which suggests implementing this host-side (set it on the Inspect generate config) rather than as a CLI passthrough. Curious what your fork observed in production, since that may point at a different bridge configuration.

Also, the PR body describes a `MAX_THINKING_TOKENS` env-var implementation with three levels, but the diff ships a `--effort` flag with five. Worth correcting either way, since the body survives the squash.

src/inspect_swe/_claude_code/claude_code.py
sjawhar: Added: allowlist_bridged_tools=False now raises a ValueError at construction unless permission_mode='auto', matching the auto_mode/permission_mode conflict check in resolve_claude_code_deprecated_args. Pushed in 82577bf1de8f.

sjawhar:
Thanks for running this end to end and tracing it through the bridge.

allowlist_bridged_tools: added the ValueError guard at construction, replied inline on that thread.

effort: you're right, and I confirmed it against the bridge source. claude_code() never sets forward_generation_config on sandbox_agent_bridge(), so the request-level generation config the CLI attaches to its own outgoing call gets dropped before generation, and only the resolved Inspect model's own config governs. The --effort CLI flag was doing nothing to the served model in any configuration this agent reaches. I reworked resolve_claude_code_models() to merge GenerateConfig(reasoning_effort=effort) onto a copy of every served model it resolves (the primary model plus any opus/sonnet/haiku/subagent role that gets its own alias), which is the host-side fix you suggested, and removed the CLI passthrough (claude_code_effort_args/claude_code_command) entirely since it had no effect in any reachable configuration. Rewrote tests/test_claude_code_effort.py against the new mechanism with mockllm, no live dependency.

On the PR body: it described the old MAX_THINKING_TOKENS/three-level approach, which didn't match the diff even before this fix (the diff always shipped --effort with five levels). I've corrected the body to describe the current GenerateConfig-based implementation, and I want to flag that the body's closing claim that this shipped and ran in production evals isn't accurate for the effort half given what you found; the CLI flag ran but had no effect, so nothing was actually exercising a different reasoning effort in those evals.

On splitting into two PRs: happy to if you'd still like that now that effort actually works and both halves are in similar shape. Let me know and I'll open the split.

epatey:
@auto

i-am-marvin:
**Claude finished @epatey's task in 11m 24s** —— [View job](https://github.com/meridianlabs-ai/inspect_swe/actions/runs/33433334326)

---
### Verified: both review asks are addressed on the current head; no changes needed

- [x] Bring branch up to date with `main` (already up to date — `d56c5be` merged it)
- [x] Verify the `allowlist_bridged_tools` ValueError guard from 82577bf
- [x] Verify the host-side `effort` implementation in `resolve_claude_code_models()`
- [x] Run lint, type, and test checks
- [x] Resolve the addressed inline review thread
- [x] Report results

### What I checked

**ValueError guard** — present at `src/inspect_swe/_claude_code/claude_code.py:271`: `allowlist_bridged_tools=False` with any effective permission mode other than `"auto"` raises at construction, including the default (no `permission_mode`) case, which resolves to `--dangerously-skip-permissions` and is correctly rejected. Both guard tests pass (`test_allowlist_bridged_tools_false_requires_permission_mode_auto`, `test_allowlist_bridged_tools_false_rejects_non_auto_permission_mode`). I resolved the inline review thread since the ask is fully addressed and verified.

**Host-side `effort`** — `resolve_claude_code_models()` (`src/inspect_swe/_claude_code/model.py:75-82`) now merges `GenerateConfig(reasoning_effort=effort)` onto a shallow copy of every served model it resolves (primary plus any per-role alias), and the CLI passthrough is gone from the diff entirely. One subtlety I verified directly: `get_model()` returns cached/shared instances, and the `copy()` correctly isolates the config override — the original cached model's `reasoning_effort` stays `None` while the copy gets the override, so `effort` can't leak into other agents sharing the same model in an eval. Caller-supplied `model_aliases` are left untouched as documented.

### Checks run locally (fresh install of the PR head)

- `ruff format --check` / `ruff check`: clean
- `mypy` (strict): no issues in 121 source files
- `tests/test_claude_code_effort.py` + `tests/test_claude_code_mcp_allowed_tools.py`: 19 passed
- `tests/test_claude_code_model.py`: 4 passed
- PR CI: all 10 checks passing on the current head

### Not done

- **PR split**: sjawhar's offer to split into two PRs is still open and is a maintainer call — epatey's `@auto` trigger didn't answer it, so I didn't act on it. Both halves are now in mergeable shape either way.
- I pushed no commits (the branch was already current with `main` and nothing needed fixing), so I have not triggered a re-review.

epatey (approved):
Thanks, this addresses everything from the earlier round: the construction-time guard, the host-side effort rework, and the body correction all look good. Keeping it as one PR since both halves are clean now. Approving.

PR #123, opened 2026-08-12
fix: make bridged MCP readiness timeouts configurable

sjawhar:
## Scope

Follow-up to #106. Adds `mcp_ready_timeout` to every gated agent entry point, including OpenCode and the public ACP base path. Each readiness wait receives the configured value.

Codex now applies a 300-second MCP startup timeout only to bridged servers. Static caller-provided servers retain Codex defaults. `mcp_startup_timeout` permits an explicit bridged-server value or `None` for the Codex default.

All affected agent option tables document the timeout settings.

## Verification

Claude ran `tests/test_mcp_ready.py` and `tests/test_codex_config.py`: 57 passed. Claude ran ruff on the changed Python paths: passed. The repository-wide mypy gate passes on CI for 3.10 and 3.11.

### DIFF
--- a/src/inspect_swe/_codex_cli/config.py
+++ b/src/inspect_swe/_codex_cli/config.py
@@ -7,6 +7,8 @@ from inspect_ai.tool import MCPServerConfig
 from pydantic import BaseModel, ConfigDict, Field
 from typing_extensions import TypedDict
 
+MCP_STARTUP_TIMEOUT_SEC = 300
+
 CodexWebSearch = Literal["live", "cached", "disabled"]
 
 
@@ -386,7 +388,9 @@ def resolve_codex_sandbox_mode(
 
 
 def codex_mcp_server_config(
-    mcp_server: MCPServerConfig, bridged_server_names: AbstractSet[str]
+    mcp_server: MCPServerConfig,
+    bridged_server_names: AbstractSet[str],
+    bridged_startup_timeout: int | None,
 ) -> dict[str, Any]:
     """TOML table for one `mcp_servers.<name>` entry in codex config.
 
@@ -402,6 +406,8 @@ def codex_mcp_server_config(
     server_config = mcp_server.model_dump(exclude={"name", "tools"}, exclude_none=True)
     if mcp_server.name in bridged_server_names:
         server_config["required"] = True
+        if bridged_startup_timeout is not None:
+            server_config["startup_timeout_sec"] = bridged_startup_timeout
     return server_config
 
 
@@ -410,6 +416,7 @@ def codex_mcp_servers_toml(
     bridged_server_names: AbstractSet[str],
     approval_policy: CodexApprovalPolicy,
     force_approve: bool,
+    bridged_startup_timeout: int | None,
 ) -> dict[str, dict[str, Any]]:
     """Build `[mcp_servers.<name>]` TOML tables for one class of MCP server.
 
@@ -423,11 +430,15 @@ def codex_mcp_servers_toml(
     return {
         f"mcp_servers.{mcp_server.name}": (
             codex_mcp_server_toml(
-                codex_mcp_server_config(mcp_server, bridged_server_names),
+                codex_mcp_server_config(
+                    mcp_server, bridged_server_names, bridged_startup_timeout
+                ),
                 approval_policy,
             )
             if force_approve
-            else codex_mcp_server_config(mcp_server, bridged_server_names)
+            else codex_mcp_server_config(
+                mcp_server, bridged_server_names, bridged_startup_timeout
+            )
         )
         for mcp_server in mcp_servers
     }
--- a/src/inspect_swe/_gemini_cli/gemini_cli.py
+++ b/src/inspect_swe/_gemini_cli/gemini_cli.py
@@ -23,7 +23,10 @@ from inspect_ai.util._sandbox import ExecRemoteAwaitableOptions
 
 from inspect_swe._util._async import is_callable_coroutine
 from inspect_swe._util.centaur import CentaurOptions, run_centaur
-from inspect_swe._util.mcp_ready import wait_for_mcp_endpoints
+from inspect_swe._util.mcp_ready import (
+    DEFAULT_MCP_READY_TIMEOUT,
+    wait_for_mcp_endpoints,
+)
 from inspect_swe._util.messages import build_user_prompt
 from inspect_swe._util.path import join_path
 from inspect_swe._util.sandbox import resolve_agent_cwd
@@ -43,6 +46,7 @@ def gemini_cli(
     skills: Sequence[str | Path | Skill] | None = None,
     mcp_servers: Sequence[MCPServerConfig] | None = None,
     bridged_tools: Sequence[BridgedToolsSpec] | None = None,
+    mcp_ready_timeout: float = DEFAULT_MCP_READY_TIMEOUT,
     web_search: bool = True,
     centaur: bool | CentaurOptions = False,
     attempts: int | AgentAttempts = 1,
@@ -73,6 +77,8 @@ def gemini_cli(
         skills: Additional [skills](https://inspect.aisi.org.uk/tools-standard.html#sec-skill) to make available to the agent.
         mcp_servers: MCP servers to make available to the agent
         bridged_tools: Host-side Inspect tools to expose to the agent via MCP
+        mcp_ready_timeout: Seconds to wait for bridged MCP endpoints to serve
+            tools before the agent launch errors.
         web_search: Enable the agent's web search tool (defaults to `True`).
         centaur: Run in 'centaur' mode, which makes Gemini CLI available to an Inspect `human_cli()` agent rather than running it unattended.
         attempts: Configure agent to make multiple attempts
@@ -218,7 +224,11 @@ def gemini_cli(
             ]
             if _http_mcp_configs:
                 await wait_for_mcp_endpoints(
-                    _http_mcp_configs, bridge, sandbox=sandbox, required=True
+                    _http_mcp_configs,
+                    bridge,
+                    sandbox=sandbox,
+                    timeout=mcp_ready_timeout,
+                    required=True,
                 )
 
             if centaur:
@@ -250,7 +260,11 @@ def gemini_cli(
                     # iteration.
                     if _http_mcp_configs and attempt_count > 0:
                         await wait_for_mcp_endpoints(
-                            _http_mcp_configs, bridge, sandbox=sandbox, required=True
+                            _http_mcp_configs,
+                            bridge,
+                            sandbox=sandbox,
+                            timeout=mcp_ready_timeout,
+                            required=True,
                         )
                     result = await sbox.exec_remote(
                         cmd=["bash", "-c", 'exec 0</dev/null; "$@"', "bash"]
--- a/src/inspect_swe/acp/agent.py
+++ b/src/inspect_swe/acp/agent.py
@@ -21,7 +21,10 @@ from inspect_ai.tool import MCPServerConfig, MCPServerConfigHTTP
 from inspect_ai.util import ExecRemoteProcess
 from typing_extensions import TypedDict, Unpack
 
-from inspect_swe._util.mcp_ready import DEFAULT_MCP_READY_TIMEOUT, wait_for_mcp_endpoints
+from inspect_swe._util.mcp_ready import (
+    DEFAULT_MCP_READY_TIMEOUT,
+    wait_for_mcp_endpoints,
+)
 
 from .client import ACPError, acp_connection, format_acp_failure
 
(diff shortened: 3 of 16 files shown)


tests/test_mcp_ready.py
```diff
@@ -0,0 +1,813 @@
...
+    root = Path(__file__).parent.parent / "src" / "inspect_swe"
+    # antigravity belongs here for the same reason as the others: it gates its
+    # SDK launch on wait_for_mcp_endpoints, so a non-configurable timeout pins
+    # it to the default. It was omitted when this list was written, and the
+    # omission cost a 720-sample eval-set in which EVERY recorded sample died
+    # with MCPEndpointsUnreachableError after 120s -- the bridge answers in
+    # ~1s at one sample, but not while ~400 sandboxes share the runner's
+    # single event loop. Adding an agent to this dict is what makes that
+    # failure mode impossible to reintroduce silently.
+    agent_paths = {
```
craig-meridian: [P2] Discover the gated call sites instead of hand-listing them. This dict is the same omission mechanism the comment above it memorializes: a seventh gated agent (or a new gate added to a module not in this list) can await `wait_for_mcp_endpoints` without forwarding a configurable timeout and the suite stays green until somebody remembers to update the dictionary — the pattern that previously missed OpenCode and ACP.

The right shape already exists two tests up in this same file: `test_every_self_launching_bridged_agent_gates_on_mcp_readiness` and the centaur-gate guard both `rglob` the package and predicate on AST nodes. Auto-discovering every awaited `wait_for_mcp_endpoints` call under `src/inspect_swe` (excluding `_util/mcp_ready.py`) currently finds exactly the seven covered files, and asserting each discovered call passes `timeout=` as an `ast.Name` **or** `ast.Attribute` also subsumes `test_acp_agent_forwards_configured_mcp_readiness_timeout` — whose `assert len(waits) == 1` would fail spuriously if ACP ever grows a second, correctly-forwarding gate. (Keeping ACP as a separate structural case for the TypedDict half is fine; the call-site half should be discovered.)

Related drift worth fixing in the same pass: these new matchers accept only `ast.Name` callees, while the PR-106 helpers in this file (`has_awaited_gate` and friends) also accept `ast.Attribute`. A module-qualified `mcp_ready.wait_for_mcp_endpoints(...)` would still satisfy the older guards but make `readiness_calls` empty here, failing `assert readiness_calls` with no hint why. One shared module-level matcher used by all of these tests removes the drift.

tests/test_mcp_ready.py
```diff
@@ -0,0 +1,813 @@
...
+            node
+            for node in tree.body
+            if isinstance(node, ast.FunctionDef) and node.name == agent_name
+        )
+        args = function.args.args
+        timeout_index = next(
+            index for index, arg in enumerate(args) if arg.arg == "mcp_ready_timeout"
+        )
+        first_default_index = len(args) - len(function.args.defaults)
+        default = function.args.defaults[timeout_index - first_default_index]
```
craig-meridian: [P3] This defaults arithmetic has two holes, both demonstrated by running the test's exact extraction logic against modified signatures:

1. `timeout_index - first_default_index` can go negative, and Python's negative indexing then silently reads *another parameter's* default. With `def agent(mcp_ready_timeout, name="x", other=DEFAULT_MCP_READY_TIMEOUT)` the guard **passes** even though `mcp_ready_timeout` has no default at all — the exact property this assertion exists to pin.
2. If these factories ever adopt keyword-only parameters (per the positional-compatibility discussion), the parameter moves to `args.kwonlyargs`/`kw_defaults`, and the `next(...)` above dies with a bare `StopIteration` that reads as a broken test rather than a regression.

A small helper closes both and gives every failure path a message naming the function:

```python
def _param_default(fn: ast.FunctionDef, name: str) -> ast.expr:
    offset = len(fn.args.args) - len(fn.args.defaults)
    for i, arg in enumerate(fn.args.args):
        if arg.arg == name:
            assert i >= offset, f"{fn.name}: {name} has no default"
            return fn.args.defaults[i - offset]
    for arg, default in zip(fn.args.kwonlyargs, fn.args.kw_defaults):
        if arg.arg == name:
            assert default is not None, f"{fn.name}: {name} has no default"
            return default
    raise AssertionError(f"{fn.name}() has no parameter named {name!r}")
```

Naturally lands as part of the auto-discovery rewrite above — same lines of the same test.

src/inspect_swe/_gemini_cli/gemini_cli.py
```diff
@@ -43,6 +47,7 @@ def gemini_cli(
     mcp_servers: Sequence[MCPServerConfig] | None = None,
     bridged_tools: Sequence[BridgedToolsSpec] | None = None,
     web_search: bool = True,
+    mcp_ready_timeout: float = DEFAULT_MCP_READY_TIMEOUT,
```
craig-meridian: [nit] gemini_cli is the only factory that places `mcp_ready_timeout` after `web_search`; the other five put it immediately after `bridged_tools`, and the gemini docs table lists it before `web_search`, so the signature disagrees with both its siblings and its own documentation row. Please move it up one slot to match.

src/inspect_swe/_codex_cli/config.py
```diff
@@ -157,3 +161,27 @@ def check_codex_auto_review_version(version: str | None) -> None:
             f"(found {version}). Pass version='latest' (or an explicit newer "
             "version) to codex_cli()."
         )
+
+
+def codex_mcp_server_config(
+    mcp_server: MCPServerConfig,
+    bridged_server_names: AbstractSet[str],
+    bridged_startup_timeout: int | None = MCP_STARTUP_TIMEOUT_SEC,
```
craig-meridian: [P3] This default is dead in production — the only production caller (codex_cli.py:346) always passes `mcp_startup_timeout` explicitly — which leaves the factory wiring untested: drop the third argument at that call site and the whole suite stays green (these unit tests exercise this function's own default, which is the same constant), while `codex_cli(mcp_startup_timeout=600)` and `mcp_startup_timeout=None` silently become no-ops. Making the parameter required turns that wiring failure into a loud TypeError and removes the duplicated default; the existing two-arg test calls just gain an explicit argument.

src/inspect_swe/acp/agent.py
```diff
@@ -107,6 +115,9 @@ def __init__(self, **kwargs: Unpack[ACPAgentParams]) -> None:
         self.env: dict[str, str] = kwargs.get("env") or {}
         self.user = kwargs.get("user")
         self.sandbox = kwargs.get("sandbox")
+        self.mcp_ready_timeout = kwargs.get(
```
craig-meridian: [nit, non-blocking] Unlike the sibling params in this `__init__` (`env`, `model`, etc. use `or`-fallbacks), an explicit `mcp_ready_timeout=None` is stored verbatim here and crashes at agent-launch time inside the deadline arithmetic (`deadline = start + timeout` in mcp_ready.py) with an unrelated-looking TypeError. Since `mcp_startup_timeout=None` legitimately means "use the Codex default" one parameter over in codex_cli, a `None` here is a plausible caller mistake. Suggest resolving it explicitly (`or` would wrongly swallow `0`):

```python
timeout = kwargs.get("mcp_ready_timeout")
self.mcp_ready_timeout = DEFAULT_MCP_READY_TIMEOUT if timeout is None else timeout
```

tests/test_mcp_ready.py
craig-meridian: [P2] Discover the gated call sites instead of hand-listing them. This dict is the same omission mechanism the comment above it memorializes: a seventh gated agent (or a new gate added to a module not in this list) can await `wait_for_mcp_endpoints` without forwarding a configurable timeout and the suite stays green until somebody remembers to update the dictionary — the pattern that previously missed OpenCode and ACP.

The right shape already exists two tests up in this same file: `test_every_self_launching_bridged_agent_gates_on_mcp_readiness` and the centaur-gate guard both `rglob` the package and predicate on AST nodes. Auto-discovering every awaited `wait_for_mcp_endpoints` call under `src/inspect_swe` (excluding `_util/mcp_ready.py`) currently finds exactly the seven covered files, and asserting each discovered call passes `timeout=` as an `ast.Name` **or** `ast.Attribute` also subsumes `test_acp_agent_forwards_configured_mcp_readiness_timeout` — whose `assert len(waits) == 1` would fail spuriously if ACP ever grows a second, correctly-forwarding gate. (Keeping ACP as a separate structural case for the TypedDict half is fine; the call-site half should be discovered.)

Related drift worth fixing in the same pass: these new matchers accept only `ast.Name` callees, while the PR-106 helpers in this file (`has_awaited_gate` and friends) also accept `ast.Attribute`. A module-qualified `mcp_ready.wait_for_mcp_endpoints(...)` would still satisfy the older guards but make `readiness_calls` empty here, failing `assert readiness_calls` with no hint why. One shared module-level matcher used by all of these tests removes the drift.

tests/test_mcp_ready.py
craig-meridian: [P3] This defaults arithmetic has two holes, both demonstrated by running the test's exact extraction logic against modified signatures:

1. `timeout_index - first_default_index` can go negative, and Python's negative indexing then silently reads *another parameter's* default. With `def agent(mcp_ready_timeout, name="x", other=DEFAULT_MCP_READY_TIMEOUT)` the guard **passes** even though `mcp_ready_timeout` has no default at all — the exact property this assertion exists to pin.
2. If these factories ever adopt keyword-only parameters (a plausible future refactor for signatures this long), the parameter moves to `args.kwonlyargs`/`kw_defaults`, and the `next(...)` above dies with a bare `StopIteration` that reads as a broken test rather than a regression.

A small helper closes both and gives every failure path a message naming the function:

```python
def _param_default(fn: ast.FunctionDef, name: str) -> ast.expr:
    offset = len(fn.args.args) - len(fn.args.defaults)
    for i, arg in enumerate(fn.args.args):
        if arg.arg == name:
            assert i >= offset, f"{fn.name}: {name} has no default"
            return fn.args.defaults[i - offset]
    for arg, default in zip(fn.args.kwonlyargs, fn.args.kw_defaults):
        if arg.arg == name:
            assert default is not None, f"{fn.name}: {name} has no default"
            return default
    raise AssertionError(f"{fn.name}() has no parameter named {name!r}")
```

Naturally lands as part of the auto-discovery rewrite above — same lines of the same test.

src/inspect_swe/_gemini_cli/gemini_cli.py
craig-meridian, line 49: [nit] gemini_cli is the only factory that places `mcp_ready_timeout` after `web_search`; the other five put it immediately after `bridged_tools`, and the gemini docs table lists it before `web_search`, so the signature disagrees with both its siblings and its own documentation row. Please move it up one slot to match.

src/inspect_swe/_codex_cli/config.py
craig-meridian: [P3] This default is dead in production — the only production caller (codex_cli.py:346) always passes `mcp_startup_timeout` explicitly — which leaves the factory wiring untested: drop the third argument at that call site and the whole suite stays green (these unit tests exercise this function's own default, which is the same constant), while `codex_cli(mcp_startup_timeout=600)` and `mcp_startup_timeout=None` silently become no-ops. Making the parameter required turns that wiring failure into a loud TypeError and removes the duplicated default; the existing two-arg test calls just gain an explicit argument.

src/inspect_swe/acp/agent.py
craig-meridian: [nit, non-blocking] Unlike the sibling params in this `__init__` (`env`, `model`, etc. use `or`-fallbacks), an explicit `mcp_ready_timeout=None` is stored verbatim here and crashes at agent-launch time inside the deadline arithmetic (`deadline = start + timeout` in mcp_ready.py) with an unrelated-looking TypeError. Since `mcp_startup_timeout=None` legitimately means "use the Codex default" one parameter over in codex_cli, a `None` here is a plausible caller mistake. Suggest resolving it explicitly (`or` would wrongly swallow `0`):

```python
timeout = kwargs.get("mcp_ready_timeout")
self.mcp_ready_timeout = DEFAULT_MCP_READY_TIMEOUT if timeout is None else timeout
```

craig-meridian (changes requested):
Thanks Sami — the three findings carried over from #106 are all properly addressed: `mcp_ready_timeout` reaches OpenCode and the public ACP path (every awaited `wait_for_mcp_endpoints` in the package now forwards it, and none of the rewritten call sites lost `sandbox=`/`required=True`), the Codex 300-second startup timeout applies only to bridged servers with bridged/static/override coverage, and all six option tables document the new settings. I ran the three test files locally (64 passed) plus ruff and mypy (clean); CI is green.

Two housekeeping items now that #106 is merged:

- Please rebase onto main so this PR carries only the two follow-up commits (the branch still contains the pre-squash #106 commits, so the diff and CI base are stale).
- Please refresh the Verification section of the PR body: repository-wide mypy passes on CI for 3.10 and 3.11 (and locally against the frozen lockfile), so the sentence about nine upstream `web_search` errors is stale.

craig-meridian:
Status: recent commits are rebase/housekeeping only — review comments still open, awaiting author.

craig-meridian (approved):
Thanks Sami, approved.

PR #124, opened 2026-08-16
feat: expose agent binary installation APIs

sjawhar:
## Summary
- Expose the async agent binary installer and Claude Code/Codex source factories from inspect_swe.
- Add a public-import regression test.

## Verification
A fresh reviewer imported ensure_agent_binary_installed, claude_code_binary_source, and codex_cli_binary_source from inspect_swe and confirmed each is callable. The focused Codex agent-binary test suite passes (9 tests).

### DIFF
--- a/src/inspect_swe/__init__.py
+++ b/src/inspect_swe/__init__.py
@@ -1,5 +1,7 @@
 from ._antigravity.antigravity import antigravity
+from ._claude_code.agentbinary import claude_code_binary_source
 from ._claude_code.claude_code import claude_code
+from ._codex_cli.agentbinary import codex_cli_binary_source
 from ._codex_cli.codex_cli import codex_cli
 from ._codex_cli.config import CodexAutoReview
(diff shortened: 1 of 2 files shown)


dragonstyle:
Can you please some explanation why exposing this is useful? Want to understand the use base before we extend the public API.

sjawhar:
Closing this — the need went away.

These exports were a by-product of separate work to install the Codex code-mode host, where we resolved `codex-code-mode-host-{target}.tar.gz` as an extra release asset and wanted the installer/source factories reachable from a consumer. #107 has since made that unnecessary: preferring the full `codex-package-{arch}.tar.gz` archive already ships the code-mode host, so we no longer need our own install path, and nothing of ours calls these three symbols.

Sorry for the noise — the PR should have stated the motivation up front, and by the time you asked, the motivation had already been superseded by your own change. Thanks for asking rather than just merging it.

Issue #125, opened 2026-08-16
Investigate broken slow tests


dragonstyle:
@auto please investigate this failed run and open a pr to fix if necessary.

https://github.com/meridianlabs-ai/actions/actions/runs/31935797486

i-am-marvin:
**Claude finished @dragonstyle's task in 5m 36s** —— [View job](https://github.com/meridianlabs-ai/inspect_swe/actions/runs/31957780621) • [`claude/issue-125-20260816-1609`](https://github.com/meridianlabs-ai/inspect_swe/tree/claude/issue-125-20260816-1609) • [Create PR ➔](https://github.com/meridianlabs-ai/inspect_swe/compare/main...claude/issue-125-20260816-1609?quick_pull=1&title=fix%3A%20send%20GITHUB_TOKEN%20on%20GitHub%20API%20requests%20to%20avoid%20rate%20limits&body=Nightly%20slow%20tests%20(meridianlabs-ai%2Factions%20run%2031935797486)%20failed%2019%20tests%20because%20unauthenticated%20requests%20to%20api.github.com%20(codex%2Fgemini%2Fopencode%20release%20resolution)%20hit%20GitHub%27s%2060%2Fhour%20per-IP%20rate%20limit%2C%20which%20shared%20CI%20runner%20IPs%20routinely%20exhaust.%0A%0AThis%20attaches%20a%20Bearer%20token%20from%20GITHUB_TOKEN%20(or%20GH_TOKEN)%20to%20requests%20scoped%20to%20api.github.com%20only%2C%20raising%20the%20limit%20to%205%2C000%2Fhour.%20Adds%20unit%20tests%20for%20token%20pickup%2C%20GH_TOKEN%20fallback%2C%20and%20host%20scoping.%0A%0ANote%3A%20the%20nightly%20workflow%20in%20meridianlabs-ai%2Factions%20also%20needs%20%60GITHUB_TOKEN%3A%20%24%7B%7B%20github.token%20%7D%7D%60%20added%20to%20the%20test%20step%20env%20for%20this%20to%20take%20effect%20there.%0A%0ACloses%20%23125%0A%0AGenerated%20with%20%5BClaude%20Code%5D(https%3A%2F%2Fclaude.ai%2Fcode))

---
### Investigated broken slow tests — fix pushed

- [x] Fetch logs from the failed run
- [x] Identify the failure cause
- [x] Determine if a fix is needed in this repo
- [x] Implement fix
[...]

PR #127, opened 2026-08-16
fix: resolve npm agent versions once per process to avoid GitHub rate limits

dragonstyle:
Alternative to #126 for the nightly failures in #125, fixing the cause rather than raising the ceiling.

## The problem

`gemini_cli` and `opencode` resolve their version by calling `api.github.com/repos/.../releases/latest` **once per sample**:

- `_gemini_cli/agentbinary.py` → `resolve_gemini_version()` → `_fetch_latest_release()`
- `_opencode/agentbinary.py` → `resolve_opencode_version()` → `_fetch_latest_release()`

Both are called from inside the per-sample agent body (`_gemini_cli/gemini_cli.py:139`, `_opencode/opencode.py:143`). Unlike agents installed through `AgentBinarySource`, they never pass through the `_resolved_versions` memo in `download_agent_binary_async`, because they install from npm rather than a release asset. The `concurrency()` guard sits downstream in `ensure_*_installed`, so it doesn't deduplicate the fetch either.

Unauthenticated GitHub API requests are limited to 60/hour per IP. A single 60-sample eval can therefore exhaust the limit by itself, failing with a bare `403 Forbidden` — there is no retry, and the offline fallback in `ensure_agent_binary_installed` only covers pinned versions.

This affects local users as much as CI; it is not a shared-runner-IP problem.

## The change

Adds `_util/versioncache.py` and routes both resolvers through it.

- 60 requests per eval → 1.
- Resolution is **single-flight**: `concurrency(1)` per key with a re-check after acquiring, matching the double-checked pattern already used by `codex_models_catalog`. A caching memo alone isn't enough here — the cache dict is guarded by a `threading.Lock` that isn't held across the resolve `await`, so a cold-start burst of samples would otherwise all miss and each issue a request. The `AgentBinarySource` memo this was modelled on avoids that only because codex resolves *inside* the per-binary `concurrency(1)` guard; gemini and opencode resolve outside it, which is the bug being fixed.
- `concurrency()` is inspect's own async primitive, so unlike a module-level `anyio.Lock` it holds no waiter state bound to a particular event loop — which matters because every `anyio.run()` call creates and closes a fresh one.
- All four aliases (`auto`/`sandbox`/`stable`/`latest`) share one cache entry, since they resolve to the same release.
- An explicit semver still short-circuits with no request at all, unchanged.
- Failed resolutions are not cached, so a later caller retries.
- No credentials involved, and no change to what gets installed.

`codex_cli` is unaffected — it already resolves through the existing memo at 2 requests per process.

## Tests

12 tests in `tests/test_version_cache.py` covering the cache utility (resolve-once, key independence) and, for both agents: repeated `latest` resolution issuing one fetch, alias sharing, explicit versions never fetching, failures not being cached, and ten concurrent callers producing exactly one request. That last one was verified to fail without the single-flight guard (10 requests instead of 1), so it's a real regression test rather than a tautology.

Full suite: 198 passed, 65 skipped. ruff and mypy clean.

## Not included

Two follow-ups discussed but deliberately left out to keep this focused:

- Resolving these two agents from the npm registry instead of GitHub releases — they install from npm anyway, so the current split can resolve a version npm cannot yet serve.
- Pinning sha256 digests for the node, ripgrep, and uv downloads, which are currently written into the sandbox and `chmod +x`'d with no verification, and whose local caches are read back unverified.

Regarding #126: with this merged, codex's 2 requests per process are the only remaining API usage, so the token is no longer needed for the rate limit itself. If you keep it as insurance against a drained shared runner IP, consider a dedicated `INSPECT_SWE_GITHUB_TOKEN` rather than reading ambient `GITHUB_TOKEN`/`GH_TOKEN` — `secrets.GITHUB_TOKEN` has to be plumbed into the workflow `env:` explicitly either way, and Codespaces exports `GITHUB_TOKEN` automatically, so ambient pickup would use a developer's PAT without their knowledge.

### DIFF
--- a/src/inspect_swe/_gemini_cli/agentbinary.py
+++ b/src/inspect_swe/_gemini_cli/agentbinary.py
@@ -15,6 +15,7 @@ from .._util.sandbox import (
     bash_command,
     detect_sandbox_platform,
 )
+from .._util.versioncache import cached_version_resolution
 
 
 async def ensure_gemini_cli_setup(
@@ -40,10 +41,17 @@ async def resolve_gemini_version(
 ) -> str:
     """Resolve version string to an actual semver version."""
     if version in ["auto", "sandbox", "stable", "latest"]:
-        release = await _fetch_latest_release()
-        return str(release["tag_name"]).lstrip("v")
+        # cached so concurrent samples don't each hit the (rate-limited)
+        # GitHub API — all four aliases resolve to the same latest release
+        return await cached_version_resolution("gemini-cli", _fetch_latest_version)
 
     return version
+
+
+async def _fetch_latest_version() -> str:
+    """Fetch the latest released gemini-cli version from GitHub."""
+    release = await _fetch_latest_release()
+    return str(release["tag_name"]).lstrip("v")
 
 
 async def ensure_gemini_cli_installed(
--- a/src/inspect_swe/_opencode/agentbinary.py
+++ b/src/inspect_swe/_opencode/agentbinary.py
@@ -16,6 +16,7 @@ from .._util.sandbox import (
     bash_command,
     detect_sandbox_platform,
 )
+from .._util.versioncache import cached_version_resolution
 
 
 async def ensure_opencode_setup(
@@ -44,10 +45,17 @@ async def resolve_opencode_version(
 ) -> str:
     """Resolve version string to an actual semver version."""
     if version in ["auto", "sandbox", "stable", "latest"]:
-        release = await _fetch_latest_release()
-        return str(release["tag_name"]).lstrip("v")
+        # cached so concurrent samples don't each hit the (rate-limited)
+        # GitHub API — all four aliases resolve to the same latest release
+        return await cached_version_resolution("opencode", _fetch_latest_version)
 
     return version
+
+
+async def _fetch_latest_version() -> str:
+    """Fetch the latest released opencode version from GitHub."""
+    release = await _fetch_latest_release()
+    return str(release["tag_name"]).lstrip("v")
 
 
 async def ensure_opencode_installed(
--- a/src/inspect_swe/_util/versioncache.py
+++ b/src/inspect_swe/_util/versioncache.py
@@ -14,10 +14,14 @@ from typing import Awaitable, Callable
 
 from inspect_ai.util import concurrency
 
-# Guards the cache dict. Held only across synchronous dict access, never an
+# Guards the cache dicts. Held only across synchronous dict access, never an
 # await.
 _resolve_lock = threading.Lock()
 _resolved_versions: dict[str, str] = {}
+# per key: (number of failed resolutions so far, exception from the latest
+# one). Lets callers already queued behind a failing resolution share its
+# exception instead of each retrying in turn.
+_failed_resolutions: dict[str, tuple[int, Exception]] = {}
 
 
 async def cached_version_resolution(
@@ -26,8 +30,10 @@ async def cached_version_resolution(
     """Resolve a version, reusing the result for the process lifetime.
 
     Concurrent callers for the same key share a single resolution rather than
-    each issuing their own request. Failures are not cached, so a later caller
-    retries.
+    each issuing their own request — including its failure, so a burst of
+    callers hitting a rate-limited API produces one error, not one per caller.
+    Failures are not cached: a call arriving after a failed resolution has
+    completed retries.
 
     Args:
         key: Cache key identifying what is being resolved (e.g. ``"opencode"``).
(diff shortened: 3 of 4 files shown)


dragonstyle:
@i-am-marvin please review the comments and address them

i-am-marvin:
**Claude finished @dragonstyle's task in 28m 6s** —— [View job](https://github.com/meridianlabs-ai/inspect_swe/actions/runs/31962597521)

---
### Review comments addressed

- [x] Gather context: read review comments and current code
- [x] Bring branch up to date with `origin/main` (already up to date — merge-base equals `origin/main` HEAD, no merge needed)
- [x] Verify/address: single-flight protection in `versioncache.py`
- [x] Verify/address: suspending mock in `test_version_cache.py` concurrency test
- [x] Run tests, ruff, mypy
- [x] Resolve addressed review threads
- [x] Push any changes and post summary (no changes needed — nothing to push)

### Summary

Both inline review comments were already addressed by the commits pushed after the review (`61ab0ac` "make version resolution single-flight across concurrent samples" and `5a49089` "trim comments to what the code needs"). I verified each against the current code rather than taking the commit messages on faith, then resolved the two threads:

**1. Single-flight protection in `versioncache.py` (thread on line 44 — resolved).** The miss path now acquires `concurrency(f"version-resolution-{key}", 1, visible=False)` and re-checks the cache after acquiring, exactly the double-checked pattern the review suggested (`src/inspect_swe/_util/versioncache.py:39-54`). The docstring's inapplicable "merely costs one extra request" phrasing is also gone.

**2. Suspending mock in the concurrency test (thread on line 134 — resolved).** `test_concurrent_resolution_makes_one_request` now uses a plain async fetch containing `await anyio.sleep(0.05)` so the 10 tasks genuinely overlap, and asserts exactly one fetch plus 10 correct results (`tests/test_version_cache.py:139-159`). I re-verified it's a real regression guard: with the `concurrency()` guard monkeypatched to a no-op, the same scenario produces **10** fetches, so the test would fail against the unprotected code.

### Verification

- `tests/test_version_cache.py`: 12 passed.
- ruff check and format: clean. mypy: clean (116 files).
- Full suite: 202 passed, 46 skipped, 15 failed. All 15 failures are live-agent integration tests (`claude_code`, `opencode`, `kimi_code`, `mini_swe_agent` evals in docker). I dug into the eval logs: the agents install and start fine and then fail with `"Could not resolve authentication method. Expected one of api_key, auth_token, or credentials to be set"` — missing model-API credentials on this runner, not this PR. The opencode failure occurs *after* version resolution and npm install succeed, and `claude_code` (untouched by this PR) fails identically.

No code changes were required, so nothing was committed or pushed; the branch already includes the latest `main`.

dragonstyle:
@i-am-marvin can you address this:

failures aren’t single-flight. If resolve() raises at [versioncache.py lines 45–53](https://github.com/meridianlabs-ai/inspect_swe/blob/5a4908966d8c3954952fd14218ca28bf713e0585/src/inspect_swe/_util/versioncache.py#L45-L53), every queued caller retries sequentially. I reproduced 10 concurrent callers receiving a persistent 403: 10 fetches, 10 errors. Current waiters should share the same result or exception; a genuinely later call can still retry.

i-am-marvin:
**Claude finished @dragonstyle's task in 16m 25s** —— [View job](https://github.com/meridianlabs-ai/inspect_swe/actions/runs/31964607245)

---
### Failure single-flight addressed

- [x] Gather context: read `versioncache.py` and tests
- [x] Bring branch up to date with `origin/main` (already up to date — no merge needed)
- [x] Share failures with queued waiters (genuinely later calls still retry)
- [x] Add regression test: 10 concurrent callers + persistent failure → 1 fetch, 10 shared errors
- [x] Run tests, ruff, mypy
- [x] Commit and push (`f917dfe`)

### What changed

You're right that failures weren't single-flight: only *successes* were shared, so 10 callers queued behind a failing resolution each retried in turn — exactly the herd this cache exists to prevent, in the very scenario (a persistent 403) where it matters most.

The fix in `src/inspect_swe/_util/versioncache.py` adds a per-key `_failed_resolutions` record of `(number of failed resolutions so far, exception from the latest one)`:

- Each caller snapshots the failure count for its key **before** entering the `concurrency()` queue.
- After acquiring the lock and missing the success cache, if the failure count has increased since the caller's arrival, a resolution it was effectively queued behind has failed — it re-raises that exception instead of retrying.
- A call arriving **after** the failure completed snapshots the already-incremented count, so the "count increased" check is false and it retries. Failures are still never cached as results.

The count-based comparison rather than a simple "last error" slot is what distinguishes queued waiters from genuinely later calls, and it stays correct across successive failed rounds (the resolver only runs when the stored count equals its arrival snapshot, so the `+1` on failure is a true increment). All state stays in plain dicts guarded by the existing `threading.Lock` — nothing bound to an event loop, preserving the multiple-`anyio.run()` safety the module already relies on.

### Verification

- Reproduced your scenario first: with failure-sharing neutralized (equivalent to the previous code), 10 concurrent callers with a persistently failing fetch made **10** fetches. With the fix: **1** fetch, all 10 callers receiving the same exception instance.
- New `test_concurrent_failure_is_shared` (parametrized over gemini and opencode) asserts exactly that, plus that a subsequent call retries and succeeds — so it's a real regression guard, not a tautology.
- `tests/test_version_cache.py`: 14 passed, including the pre-existing `test_failed_resolution_is_not_cached` confirming sequential retry-after-failure still works.
- Non-live suite (everything except docker/live-agent integration tests): 187 passed, 10 skipped. The live evals need model-API credentials this runner doesn't have (established in the previous job here — `claude_code`, untouched by this PR, fails identically); the PR's own CI is the gate for those.
- ruff check, ruff format, and mypy: clean.

Pushed as `f917dfe` to `claude/pr-126-security-review-f3aljs`.

PR #133, opened 2026-08-29
fix: acquire the OpenCode binary host-side, not via in-sandbox npm

sjawhar:
## Summary

`opencode` (the coding agent, run as an `inspect_swe` solver) currently fails to
install in a network-isolated sandbox: `ensure_opencode_installed` runs `npm`
*inside* the sandbox to fetch `opencode-ai`, and that install hangs waiting on
egress to the npm registry that a locked-down sandbox will never grant.

This brings `opencode` onto the same acquisition path `claude_code` and
`codex_cli` already use — `ensure_agent_binary_installed` /
`AgentBinarySource` — because OpenCode ships a standalone, self-contained
binary per platform on its GitHub releases. The runner:

1. Probes the sandbox with `which opencode` and reuses whatever is already
   installed/overlaid there (unchanged behavior for `version="auto"`/`"sandbox"`).
2. Otherwise resolves the requested version (`stable`/`latest`/an explicit
   semver) against the GitHub Releases API, downloads the matching
   platform-specific release tarball **host-side**, verifies its sha256
   checksum, unwraps the tarball to the bare binary, and writes it into the
   sandbox.

No step requires the sandbox to reach any external host, so `opencode` now
installs correctly in a network-isolated sandbox exactly like `claude_code`
and `codex_cli` already do.

## Why not keep npm as a fallback

OpenCode's GitHub releases publish a binary for every platform
`detect_sandbox_platform` can report (glibc and musl builds per arch), so
there is no platform this repo supports that lacks a standalone asset.
Keeping the npm path alongside it would mean carrying two acquisition
mechanisms for zero coverage gain, so it is removed rather than kept as a
fallback — matching `claude_code`, which has no npm path at all.

## Changes

- `src/inspect_swe/_opencode/agentbinary.py`: replace the npm-bundle install
  (`create_npm_bundle`/`install_npm_bundle`, `resolve_opencode_version`,
  `ensure_opencode_installed`, `_run_opencode_postinstall`) with
  `opencode_binary_source()` + `ensure_agent_binary_installed`, matching the
  `codex_cli_binary_source()` / `claude_code_binary_source()` pattern. Version
  resolution (`stable`/`latest` -> GitHub Releases API), checksum extraction,
  and the sandbox `which`-probe / caching all now flow through the shared
  `_util/agentbinary.py` machinery instead of bespoke opencode-only logic.
  `node` and `ripgrep` acquisition (dependency bin dirs) are unchanged.
- `tests/test_opencode_setup.py`: drop the npm-era `test_resolve_version_literal`
  test (the function it covered no longer exists) and add unit coverage for
  `opencode_binary_source()`: platform-matched asset selection, alias
  resolution before the release fetch, the "no asset for this platform" and
  "malformed digest" error paths, and cache-path/cache-listing behavior. The
  existing docker-sandbox integration test (`test_install_opencode_in_docker_sandbox`)
  is unchanged.
- `tests/test_version_cache.py`: drop opencode from the
  `versioncache.cached_version_resolution` parametrized cases — that shared
  cache belonged to the old `resolve_opencode_version` function, which no
  longer exists; version-resolution caching for opencode is now handled by
  `_util/agentbinary.py`'s own in-process cache, the same as `claude_code` and
  `codex_cli` (neither of which was ever in this parametrization).

## Testing

```
uv run pytest tests/test_opencode_setup.py tests/test_version_cache.py -v
```

14 passed, 1 skipped (`test_install_opencode_in_docker_sandbox` skips without
a docker daemon). The 6 new/changed `test_opencode_setup.py` unit tests fail
against the pre-change implementation with `AttributeError: module
'inspect_swe._opencode.agentbinary' has no attribute 'opencode_binary_source'`
(and similarly for `package_cache_dir`), confirming they exercise the new
code path rather than passing vacuously.

### DIFF
--- a/src/inspect_swe/_opencode/agentbinary.py
+++ b/src/inspect_swe/_opencode/agentbinary.py
@@ -78,8 +78,10 @@ def opencode_binary_source() -> AgentBinarySource:
                 f"release {version}"
             )
 
-        # Extract checksum (format: "sha256:xxx")
-        digest = asset.get("digest", "")
+        # Extract checksum (format: "sha256:xxx"). GitHub serves
+        # "digest": null for assets uploaded before digest support, so
+        # normalize None to "" rather than crashing on .startswith.
+        digest = asset.get("digest") or ""
         if not digest.startswith("sha256:"):
             raise RuntimeError(f"Invalid digest format: {digest}")
         expected_checksum = digest[7:]  # Remove "sha256:" prefix
@@ -96,10 +98,10 @@ def opencode_binary_source() -> AgentBinarySource:
         )
 
     def cached_binary_path(version: str, platform: SandboxPlatform) -> Path:
-        # Legacy single-binary cache layout. No longer written (resolve_version
-        # always returns package=True now), kept only so a cache populated
-        # before opencode moved to package-archive caching still serves as an
-        # offline fallback.
+        # Never written or read in practice: resolve_version always returns
+        # package=True, and opencode never shipped a single-binary cache (the
+        # npm era cached bundles under "opencode-bundles"). Defined only
+        # because AgentBinarySource requires it.
         return cached_binary_dir / f"opencode-{version}-{platform}"
 
     def cached_package_path(version: str, platform: SandboxPlatform) -> Path:
--- a/src/inspect_swe/_opencode/agentbinary.py
+++ b/src/inspect_swe/_opencode/agentbinary.py
@@ -15,7 +15,6 @@ from .._util.download import download_text_file
 from .._util.node import ensure_node_available
 from .._util.ripgrep import ensure_ripgrep_available
 from .._util.sandbox import SandboxPlatform, detect_sandbox_platform
-from .._util.tarball import extract_tarball
 
 
 async def ensure_opencode_setup(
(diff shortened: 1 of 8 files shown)


dragonstyle:
Pushed 76e8eed with three small follow-ups so this can land without another round trip:

- `download_agent_binary()` / `cached_agent_binaries()` now accept `"opencode"`. The opencode docs page already includes the `agent_installation` snippet telling users to prefetch with `download_agent_binary("opencode", ...)`, which raised `ValueError` before this — now that opencode has an `AgentBinarySource` it routes through the same path. Added unit tests for the routing and cache listing.
- `asset.get("digest") or ""` — GitHub serves `"digest": null` for assets uploaded before digest support, and `.get("digest", "")` returns `None` in that case, which would turn the intended `RuntimeError` into an `AttributeError`.
- Reworded the `cached_binary_path` comment: opencode never wrote the single-binary cache layout (the npm era cached bundles under `opencode-bundles`), so the field exists only because `AgentBinarySource` requires it.

Otherwise this looks good. I verified the asset-name candidates and digests against the live releases (all four platforms covered, including the `-baseline-musl` suffix order), and the release tarball contains a single top-level `opencode` file, matching `package_entrypoint`. One note for the changelog: `"auto"`/`"sandbox"` now genuinely probe the sandbox, which the old npm implementation never did — that matches the documented contract, but images with a preinstalled opencode will now reuse it under the default `"auto"` instead of getting latest installed.

PR #136, opened 2026-08-30
docs: add credential boundary and public-API evolution guidance

sjawhar:
## Summary

Adds a short `## Agent Guardrails` section to AGENTS.md with two rules that are review-enforced in this repo but were not written down: the sandbox credential boundary and how public function signatures evolve without breaking callers.

## Placement

AGENTS.md here is deliberately minimal (#87), so this keeps to two bullets. The reasoning for putting them here rather than CONTRIBUTING.md is #87's own: coding agents read AGENTS.md automatically and do not read CONTRIBUTING.md, and these are the rules an agent preparing a PR needs in front of it.

## Why credential boundary

The sandbox is untrusted. By default, `claude_code()` supplies a bridge-owned dummy token for its in-sandbox model bridge. A caller can override that environment through `claude_code(env=...)`, including `ANTHROPIC_AUTH_TOKEN`; a caller-provided credential is therefore sandbox-exposed. Other real credentials needed for in-sandbox transport, such as MCP headers, must remain outside model- and tool-readable paths.

## Why public API evolution

#88 was blocked in round 1 for the unshimmed removal of `auto_mode`, which had been public since 0.2.56. The merged fix is the `ClaudeCodeDeprecatedArgs` / `resolve_claude_code_deprecated_args` / `**deprecated_args: Unpack[...]` pattern still in `claude_code.py` today. #124 was closed on the use-case gate: "Want to understand the use base before we extend the public API." This names both as the standing convention.

### Agent review
- Reviewer: Claude (same model as the authoring sessions); the final deep-review pass used fresh context.
- Passes: 4. The final pass checked the documented credential boundary against the current `claude_code(env=...)` path.
- Findings: two credential-boundary overstatements were fixed. The rule now distinguishes bridge-owned defaults from caller overrides and treats caller-provided credentials as sandbox-exposed. The heading and rule-first phrasing were also reviewed for fit with AGENTS.md's deliberately minimal scope.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -33,7 +33,7 @@ This file provides guidance to AI coding agents working in this repository.
 
 ## Agent Guardrails
 
-- **Credential boundary**: treat the evaluated sandbox as untrusted. Real model-provider credentials never enter it—only dummy keys reach model inference. Other real credentials (e.g. MCP auth headers) may enter the sandbox for transport, but must stay outside every model- and tool-readable path.
+- **Credential boundary**: treat the evaluated sandbox as untrusted. By default, bridge-owned model-provider credentials never enter it—only the dummy key used to authenticate with the in-sandbox model bridge reaches inference. A `claude_code(env=...)` caller can explicitly override its environment, including `ANTHROPIC_AUTH_TOKEN`, so caller-provided credentials must be treated as sandbox-exposed. Other real credentials (e.g. MCP auth headers) may enter the sandbox for transport, but must stay outside every model- and tool-readable path.
 - **Public API evolution**: add new parameters at the end of the signature, using `Literal` types with runtime validation for constrained options. Renaming or removing a released parameter needs a `**deprecated_args: Unpack[...]` shim; new public surface needs a concrete use case first.
 
 ## Agent Review Disclosure
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,3 +7,8 @@
 - Body lines starting with `<type>:` are parsed as extra changelog entries—don't begin description lines with a conventional-commit prefix unless that's intended
 - Never edit `CHANGELOG.md`, version numbers, or `.release-please-manifest.json`—Release Please owns them
 - See [CONTRIBUTING.md](CONTRIBUTING.md) for full guidelines
+
+## Agent Guardrails
+
+- **Credential boundary**: treat the evaluated sandbox as untrusted. Real model-provider credentials never enter it—only dummy keys reach model inference. Other real credentials (e.g. MCP auth headers) may enter the sandbox for transport, but must stay outside every model- and tool-readable path.
+- **Public API evolution**: add new parameters at the end of the signature, using `Literal` types with runtime validation for constrained options. Renaming or removing a released parameter needs a `**deprecated_args: Unpack[...]` shim; new public surface needs a concrete use case first.


sjawhar:
Pushed a correction after a thermonuclear docs-review pass (head moved to `87bd36c9`):

- **Credential-boundary bullet was overstated.** "Real credentials stay host-side in bridge code; only inert test credentials ever enter the sandbox" is false for the `antigravity()` adapter: `_mcp_server_entries()` in `antigravity.py` carries each MCP server's `headers` (including a live `Authorization` token for an authenticated server) into the `RunnerPayload`, which is written as `request.json` into the sandbox filesystem (`antigravity.py:428-433`); `sdk_runner.py:183-189` passes those same headers into the in-sandbox SDK runner. The bullet now states the invariant that's actually upheld: real model-provider credentials never enter the sandbox (only `_SANDBOX_DUMMY_API_KEY` does), while other real credentials needed for in-sandbox MCP transport may enter the filesystem/process as long as they stay outside every model- and tool-readable path — which is why `_STATE_DIRECTORY_NAME` is a sibling of, not nested in, the directory `view_file` can read.
- Removed the `(#69, #73, #85, #94)` / `(#88)` / `(#124)` citations from the AGENTS.md bullets — that's PR-body provenance, not evergreen agent guidance; the evidence stays here in the PR description.
- Renamed `## Design Principles` → `## Agent Guardrails`: narrower and accurate for what these two entries are (guardrails against a security regression and an API-compatibility break), and less likely to accrete unrelated rules given AGENTS.md's deliberately-minimal intent (#87).
- Compressed both bullets to rule-first phrasing, matching the file's existing terse voice.

PR description's `### Agent review` block updated to disclose this third pass and its finding.

epatey:
Intake notes ahead of a review. These orient whoever reviews deeply; they are not the review.

**Recommendation.** This is a maintainer taste and placement call more than a review. The two rules are accurate and the diff is a clean +5 lines, so if you want them written down it is mergeable as-is. The one thing to decide first: do you want AGENTS.md expanded at all given #87's deliberately-minimal intent, or would these live better in CONTRIBUTING.md? Sami already flags this and offers to relocate. Thorough, careful PR, thanks for the groundwork.

**What it does.** Adds an `## Agent Guardrails` section to AGENTS.md with two bullets: (a) a credential-boundary rule (the sandbox is untrusted; real model-provider creds never enter, only dummy keys; other real creds like MCP auth headers may enter for transport but must stay off every model- and tool-readable path), and (b) a public-API-evolution rule (append params, `Literal` plus runtime validation, a `**deprecated_args` shim before removing a released param, a use case required for new public surface). Docs-only, no runtime effect.

**Why.** These are conventions the repo already enforces in review but has not written down. The provenance for the rules is solid (credential boundary: #85 declined, #94 merged, #69/#73 adjacent; API evolution: #88's blocked unshimmed `auto_mode` removal, #124's use-case gate).

**Beyond the diff.** Safety fact: the bullets describe the code accurately, verified against main rather than taken from the body (the dummy-key-only invariant for model inference, and MCP auth headers confined to `.antigravity-state`, a sibling of `view_file`'s readable tree so a model that could read them cannot). No runtime risk since it is docs; the only risk is guidance drifting from code later, inherent to any codified invariant. The credential-boundary bullet is the narrowed, correct version: the earlier absolute phrasing ("only inert test credentials ever enter the sandbox") was false against the antigravity adapter, and this text fixes that.

**Open questions.**
- Keep in AGENTS.md, or move to CONTRIBUTING.md? (Your call per #87.)
- Codify these two now, or keep AGENTS.md release-mechanics only?

epatey (approved):
Approving, and let's keep these in AGENTS.md rather than moving them to CONTRIBUTING.md. The two rules are accurate and worth having written down. Thanks for the careful groundwork.

sjawhar:
@epatey, the approved PR remains pending because branch protection requires the versioned `py-* (3.10)` and `py-* (3.11)` contexts while the docs-only workflow emitted generic skipped contexts; no job failed.

PR #145, opened 2026-09-03
fix(claude_code): disable auto-memory so sandboxed agents use task-provided memory tools

dragonstyle:
Companion to #144, which fixed `tests/test_mcp.py::test_claude_code_mcp` by naming the `memory` MCP server in the example prompt after it failed in both 2026-09-03 nightlies ([08:26](https://github.com/meridianlabs-ai/actions/actions/runs/33733358898), [13:04](https://github.com/meridianlabs-ai/actions/actions/runs/33758955210)). That resolved the test. This PR addresses the agent default that made a plain "memory tools" instruction ambiguous in the first place, so real tasks that provide their own memory tooling are not affected the same way.

**Root cause.** Claude Code's system prompt has an `# auto memory` section that tells the model it has a persistent file-based memory and should write memory files with the file tools. The example task's "use the memory tools to record what you found" now resolves to that instead of the `memory` MCP server. The MCP server was connected in every run (all nine `mcp__memory__*` tools were in the request); the model simply chose the file-based interpretation its system prompt describes.

**Why it surfaced now.** inspect_ai 0.3.262 ([#4893](https://github.com/UKGovernmentBEIS/inspect_ai/pull/4893)) stopped concatenating Anthropic `system` blocks in the bridge. Before that, the concatenated block began with Claude Code's `x-anthropic-billing-header:` line, which the Anthropic API treats as metadata and drops along with the rest of the block. Token accounting on the 0.3.261 log bears this out: the system block counts as 6,322 tokens via `count_tokens`, but the first call billed only 2,089 uncached tokens beyond the cached tool prefix, i.e. roughly the user message alone. So bridged Claude Code has been running without its own system prompt (and without inspect_swe's `--append-system-prompt`) until 0.3.262. The test passed only because the model never saw the competing instruction.

**Fix.** Set `CLAUDE_CODE_DISABLE_AUTO_MEMORY=1` by default in the agent env (`_claude_code/env.py`, and the ACP adapter's `_BRIDGE_SAFE_ENV`, which builds its own env). Auto-memory targets future conversations a sandbox never has, and it costs ~3k system-prompt tokens plus the turns spent writing files. Claude Code's gate treats `1/true/yes/on` as disable and an explicit `0/false/no/off` as enable (overriding `settings.json`; verified in the 2.1.236 binary CI uses), so callers who deliberately study memory persistence opt back in with `env={"CLAUDE_CODE_DISABLE_AUTO_MEMORY": "0"}`. Documented in the `env` docstring. This is not a return to pre-0.3.262 behavior: the model now sees Claude Code's full system prompt minus one section that cannot function in a sandbox.

**Verification** (main, Claude Code stable 2.1.236, Sonnet 4.5, docker):

| inspect_ai | agent env | tool calls | result |
|---|---|---|---|
| 0.3.261 | default | Bash, Bash, create_entities, add_observations, create_relations | pass |
| 0.3.262 | default | Bash, Write, Write | fail |
| 0.3.262 | this PR | Bash, create_entities, open_nodes | pass |

Fast suite: 331 passed, 48 skipped. `ruff`, `ruff format --check`, `mypy src tests` clean.

Worth noting for the team: every bridged Claude Code eval on inspect_ai < 0.3.262 ran without Claude Code's system prompt. Other behavior shifts in upcoming nightlies may trace back to the same change.

### Agent review
- Pass 1: Claude Fable 5.1 code-reviewer subagent (fresh context, same model family as author). Findings: 0. Flagged as unverifiable the claim that an explicit falsy value overrides `settings.json`; verified by the author against the 2.1.236 binary's gate (env check precedes the settings lookup).
- Pass 2: Claude Fable 5.1 adversarial design-review subagent (fresh context), which independently verified the binary gate, the bridge diff, and the three repro logs. Findings: 9 — 6 fixed (section header is `# auto memory`, not `# Memory`; "per request" token cost softened to "cached after the first call"; "has no next session" softened to "normally"; trimmed the env.py comment and test docstring that restated the commit message; simplified the `env` Args entry; ACP Claude Code agent had the same gap and now shares `DISABLE_AUTO_MEMORY_ENV`; dropped the redundant verbatim-dict test) and 3 dismissed (a docs `env` row addition was tried and then dropped as stating what the feature already implies; also importing `BLOCKING_MCP_ENV` into the ACP agent is an unrelated refactor, its values and comments differ; exporting `TRUTHY_ENV_VALUES` only for tests matches the existing `FALSY_ENV_VALUES` precedent). Recommended keeping an env-only opt-in over a new parameter.

### DIFF
--- a/docs/claude_code.qmd
+++ b/docs/claude_code.qmd
@@ -34,7 +34,7 @@ The following options are supported for customizing the behavior of the agent:
 | `retry_refusals` | Should refusals be retried? Defaults to 3. |
 | `retry_uncaught_errors` | Should uncaught errors (unexpected crashes of Claude Code) be retried? Defaults to 3. |
 | `cwd` | Working directory for {{< meta agent_name >}} session. |
-| `env` | Environment variables to set for {{< meta agent_name >}} (applied last, so they override the defaults inspect_swe sets). |
+| `env` | Environment variables to set for {{< meta agent_name >}}. |
 | `version` | Version of {{< meta agent_name >}} to use (see [Installation](#installation) below for details) |
 
 : {tbl-colwidths=\[25,75\]}
--- a/docs/claude_code.qmd
+++ b/docs/claude_code.qmd
@@ -34,7 +34,7 @@ The following options are supported for customizing the behavior of the agent:
 | `retry_refusals` | Should refusals be retried? Defaults to 3. |
 | `retry_uncaught_errors` | Should uncaught errors (unexpected crashes of Claude Code) be retried? Defaults to 3. |
 | `cwd` | Working directory for {{< meta agent_name >}} session. |
-| `env` | Environment variables to set for {{< meta agent_name >}}. Applied last, so they override the sandbox defaults inspect_swe sets (blocking MCP startup, `CLAUDE_CODE_DISABLE_AUTO_MEMORY=1`); e.g. pass `{"CLAUDE_CODE_DISABLE_AUTO_MEMORY": "0"}` to re-enable auto-memory. |
+| `env` | Environment variables to set for {{< meta agent_name >}} (applied last, so they override the defaults inspect_swe sets). |
 | `version` | Version of {{< meta agent_name >}} to use (see [Installation](#installation) below for details) |
 
 : {tbl-colwidths=\[25,75\]}
--- a/docs/claude_code.qmd
+++ b/docs/claude_code.qmd
@@ -34,7 +34,7 @@ The following options are supported for customizing the behavior of the agent:
 | `retry_refusals` | Should refusals be retried? Defaults to 3. |
 | `retry_uncaught_errors` | Should uncaught errors (unexpected crashes of Claude Code) be retried? Defaults to 3. |
 | `cwd` | Working directory for {{< meta agent_name >}} session. |
-| `env` | Environment variables to set for {{< meta agent_name >}}. |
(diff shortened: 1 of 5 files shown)


dragonstyle:
For the "not independently verified" item in the automated review: the auto-memory gate in the Claude Code 2.1.236 linux-x64 binary (the version CI's `stable` currently resolves to), minified names as shipped:

```js
function yy(){
  if(c5())return!1;                       // in-session toggle (only ever disables)
  if(yd())return!1;                       // CLAUDE_CODE_SAFE_MODE / --safe-mode
  let e=process.env.CLAUDE_CODE_DISABLE_AUTO_MEMORY;
  if($n(e))return!1;                      // $n: ["1","true","yes","on"]  -> disabled
  if(sf(e))return!0;                      // sf: ["0","false","no","off"] -> enabled
  if(V.CLAUDE_CODE_SIMPLE)return!1;
  if(V.CLAUDE_CODE_REMOTE&&!process.env.CLAUDE_CODE_REMOTE_MEMORY_DIR&&!V.CLAUDE_COWORK_MEMORY_PATH_OVERRIDE)return!1;
  if(lRn())return!1;                      // remote model-denylist flag, default off
  let t=zo();                             // settings.json
  if(t.autoMemoryEnabled!==void 0)return t.autoMemoryEnabled;
  return!0                                // default: enabled, including in --print / sdk-cli mode
}
```

So `"1"` disables, an explicit `"0"` enables ahead of the `settings.json` lookup, and there is no interactive-only gate. End-to-end this matches the repro logs: with the variable set the `# auto memory` section is absent from the first request's system blocks and the model calls the MCP tools.

PR #147, opened 2026-09-04
feat: resolve "stable"/"latest" to the agent version that will be installed

HolyMichael:
## Motivation

When I select version "stable" or "latest" I'd like to have a way to know which versions that resolves without actually having to start an agent in a sandbox. Currently this happens at install time in the sandbox.
This would be helpful on evals I want to run with multiple samples with a fixed agent. I can at setup know which version is resolved, stamp it on my run, and then at setup time for each sample use the already derived version.

# LLM implementation details and reviews.

## Summary

Adds `resolve_agent_version(agent, version, platform="linux-x64") -> str` to the public download tools and exports it from `inspect_swe`.

- `"stable"` / `"latest"` resolve to the concrete version `agent` would install in this process; an explicit version is returned unchanged. `agent` and `platform` are validated up front against their `Literal`s, so an unknown value raises before any network access.
- Reuses the shared resolver (now `resolve_agent_binary_version`, previously underscore-private with a single consumer), so it shares the per-process resolution cache the agents themselves use (one upstream lookup per process and platform, and the same answer the agent would get).
- Sync via `run_coroutine`, matching `download_agent_binary` and `cached_agent_binaries`.
- `kimi_code` is added to the download tools' agent `Literal`s and to `cached_agent_binaries()`'s aggregate branch; it already had an `AgentBinarySource` but was missing from the dispatcher. The cache filename parser is widened to accept its hyphenated `kimi-code-` prefix, without which `cached_agent_binaries("kimi_code")` returned nothing. `gemini_cli`, `mini_swe_agent` and `antigravity` install via npm/wheels/pip and have no binary source, so they are intentionally absent.

## Checks

- `ruff format` / `ruff check`: clean
- `mypy` (strict): no issues in 124 source files
- `pytest tests/test_download.py tests/test_agentbinary_install.py`: 26 passed, 8 skipped (slow gate). New tests cover the pinned-version pass-through, rejection of an unknown agent on both the `"stable"` and pinned paths, rejection of an unknown platform, and kimi cache listing; no network-dependent test added.
- Manual: `resolve_agent_version("claude_code", "stable")` → `2.1.236`, `("kimi_code", "stable")` → `0.41.0`, `("codex_cli", "stable")` → `0.153.2` at time of writing.

### Agent review
- Reviewer: Claude Fable 5.1 subagents, each in a fresh context that had not seen the authoring conversation. 4 passes, one per revision, until a pass returned no findings.
- Pass 1 (initial diff): 6 findings — 6 fixed, 0 dismissed.
  - bug: an unknown agent with a pinned version passed straight through (validation ran after the short-circuit) → dispatch first, test added for both paths.
  - convention: `kimi_code` in the dispatcher but not in the sibling public `Literal`s / aggregate listing → widened.
  - convention: only public async function in a module of sync siblings → made sync via `run_coroutine`.
  - docstring: "would install right now" overstated given the per-process cache → "in this process".
  - docstring: `platform` default can succeed where the real platform's release is missing → noted.
  - nit: unknown-agent test only exercised the `"stable"` path → parametrized.
- Pass 2 (after pass 1 fixes): 2 findings — 2 fixed, 0 dismissed.
  - bug: the `kimi_code` widening of `cached_agent_binaries` was a no-op because the filename parser rejected the hyphenated `kimi-code-` prefix (reviewer reproduced it) → parser accepts hyphenated agent prefixes, kimi cache test added.
  - nit: docstring said the version "does not vary by platform" while the cache is keyed per platform → reworded.
- Pass 3 (after pass 2 fixes): 4 findings, no bugs — 2 fixed, 2 dismissed.
  - convention: `platform` had no runtime validation, against the Literal-with-runtime-validation rule → validated via `get_args(SandboxPlatform)`, test added.
  - nit: cross-module import of underscore-private `_resolve_agent_binary_version`, now with two consumers → underscore dropped; call sites and tests updated.
  - nit, dismissed: validating `agent` on the pass-through path builds the source and so creates the cache directory. Same as the sibling `download_agent_binary`; avoiding it would duplicate the agent list.
  - nit, dismissed: the agent `Literal` is spelled out in several signatures plus the error message. Pre-existing pattern in this module; consolidating it is a separate refactor.
- Pass 4 (after pass 3 fixes): no findings. Verified every path of the new function, the parser against every cache filename the four sources produce, rename completeness, `__all__` and docs reference, the `# type: ignore` comments under strict mypy, docstring accuracy, and commit titles. One out-of-scope observation: the docs reference Agents section does not list `kimi_code`, pre-existing.

### DIFF
--- a/docs/reference/index.qmd
+++ b/docs/reference/index.qmd
@@ -17,6 +17,7 @@ reference: inspect_swe
 
 ### download_agent_binary
 ### cached_agent_binaries
+### resolve_agent_version
 ### download_wheels_tarball
 ### AgentBinary
 ### SandboxPlatform
--- a/src/inspect_swe/__init__.py
+++ b/src/inspect_swe/__init__.py
@@ -6,7 +6,12 @@ from ._gemini_cli.gemini_cli import gemini_cli
 from ._kimi_code.kimi_code import kimi_code
 from ._mini_swe_agent.mini_swe_agent import mini_swe_agent
 from ._opencode.opencode import opencode
-from ._tools.download import AgentBinary, cached_agent_binaries, download_agent_binary
+from ._tools.download import (
+    AgentBinary,
+    cached_agent_binaries,
+    download_agent_binary,
+    resolve_agent_version,
+)
 from ._util.agentwheel import download_wheels_tarball
 from ._util.centaur import CentaurOptions
 from ._util.sandbox import SandboxPlatform
@@ -38,6 +43,7 @@ __all__ = [
     "interactive_gemini_cli",
     "download_agent_binary",
     "cached_agent_binaries",
+    "resolve_agent_version",
     "AgentBinary",
     "SandboxPlatform",
     "CentaurOptions",
--- a/src/inspect_swe/_tools/download.py
+++ b/src/inspect_swe/_tools/download.py
@@ -151,23 +151,17 @@ def resolve_agent_version(
 
     Args:
         agent: Agent whose version to resolve.
-        version: Version to resolve ("stable", "latest", or an explicit version number). "auto" and "sandbox" refer to whatever is installed in the sandbox and cannot be resolved on the host, so they raise `ValueError`.
+        version: Version to resolve. "stable" and "latest" are resolved against the agent's release manifest. Any other value is returned unchanged: an explicit version number is already concrete, and "auto" / "sandbox" refer to whatever is installed in the sandbox and cannot be resolved on the host, so they pass through for the agent function to handle at install time.
         platform: Target platform ("linux-x64", "linux-arm64", "linux-x64-musl", or "linux-arm64-musl"). The version string is the same on every platform, but resolution (and its per-process cache) is per platform and consults that platform's release manifest, which may not exist for every platform.
 
     Returns:
-        The concrete version string.
+        The concrete version string for "stable" / "latest"; otherwise `version` unchanged.
     """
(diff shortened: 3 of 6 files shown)


dragonstyle (approved):
Verified locally: tests pass, mypy and ruff are clean, and resolving "stable" returns the expected concrete version for all four agents, with the second call served from the cache. The kimi_code fix also checks out: listing its binaries raised on main, and the widened filename pattern matches every existing name the same way.

I pushed one small commit to the branch: `resolve_agent_version` now raises `ValueError` for `"auto"` and `"sandbox"` instead of returning them unchanged, since those refer to whatever is installed in the sandbox and cannot be resolved on the host.

Approving. Thanks for the fix.

dragonstyle:
One more small commit on the branch, reversing my earlier one: `"auto"` and `"sandbox"` now pass through unchanged again rather than raising. They are decided per sample at install time, so the host cannot resolve them, and returning them as-is means a task that pins with `resolve_agent_version()` behaves exactly as it would without the call instead of failing at runtime. The docstring now says which inputs are resolved and which pass through. Approval stands.

HolyMichael:
@dragonstyle Thank you, I see it's in for next release. I couldn't discern a pattern to your releases, could I have some ETA for it so I can plan around it? #128

dragonstyle:
Sure thing- we're planning a release tomorrow. We are trending towards releases on Mondays but haven't yet established that rhythm...

PR #151, opened 2026-09-07
fix(opencode): prompts containing double quotes no longer surface the session title as the answer

dragonstyle:
## Problem

Running `opencode` (1.18.29) over the agent bridge on GAIA level 1 with gpt-5.5, ~4/10 samples ended with opencode's auto-generated *session title* as the sample's final answer and transcript. Every affected prompt contained a literal `"`.

## Root cause

`opencode run` wraps a positional message containing spaces in double quotes **and backslash-escapes the double quotes inside it** (`packages/opencode/src/cli/cmd/run.ts`: `` arg.includes(" ") ? `"${arg.replace(/"/g, '\\"')}"` : arg ``). The bridge's main-thread tracking anchors on the task input; it learned to accept the plain quote-wrapped form (UKGovernmentBEIS/inspect_ai#4768) but not the escaped one, so for prompts with `"` neither the answer thread nor the title thread anchored and the longer 4-message title call won in either call order.

## Fix

Deliver the prompt on stdin instead of argv, via `exec_remote(..., ExecRemoteAwaitableOptions(input=prompt))`. opencode uses piped stdin verbatim (`resolveRunInput`), so the task input now crosses the bridge byte-for-byte and anchors exactly, independent of any bridge heuristics for scaffold decorations. `exec_remote` writes `input` to the process's stdin over the sandbox-tools RPC channel and closes it, so opencode sees the EOF it needs; nothing is staged on disk, there is no shell wrapper, and there are no argv/env size limits for long prompts. The `input` option exists in inspect_ai 0.3.255 (our minimum pin) and is sandbox-agnostic. The `--continue` retry path uses the same mechanism; centaur mode is unchanged (the human types the prompt).

**Compatibility:** `resolveRunInput` (piped stdin used verbatim when there is no positional message) first shipped in opencode v1.14.42 (anomalyco/opencode#23557, 2026-05-08). Releases from 1.0.x through 1.14.41 do `message += "\n" + stdin`, so on those the model sees a leading newline before the prompt: prompts of 20+ characters still anchor by containment and beat the title call, quote-free prompts under 20 characters do not. Only reachable with an old opencode pre-installed in the sandbox image (`version="auto"` / `"sandbox"`), since downloads always resolve to a current release. Documented in 28fb331 (docs installation page, `version` docstring, delivery comment); no runtime guard, see that commit message for why.

Companion inspect_ai PR UKGovernmentBEIS/inspect_ai#5284 (bridge side): accepts the escaped rendering for users who run opencode outside inspect_swe, and fixes a second, independent bug where gpt-5.5 tool calls were dropped by opencode (`function_call` items had `id: null`, which the Vercel AI SDK rejects) so the agent stopped after its first tool call with an empty answer. That second bug accounts for the "empty after 1 tool call" samples in the same report and is **not** fixable from inspect_swe.

## Verification

Live runs (docker, gpt-5.5, opencode 1.18.29, inspect_ai main) on the exact 10 GAIA level 1 validation samples from the reported log:

| configuration | accuracy | final state |
|---|---|---|
| baseline (inspect_swe main + inspect_ai main) | 0.1 | 4 title-captured, 5 empty after first tool call, 1 correct |
| this PR only (inspect_ai main unfixed), 2-sample stand-in | | title capture fixed; tool-call drop remains |
| this PR (staged-file version) + companion inspect_ai fix | 0.7 | all 10 surface the real agent thread (3 genuine wrong answers) |
| this PR (final `exec_remote(input=...)` version) + companion inspect_ai fix, fresh run | 0.7 | all 10 surface the real agent thread, prompt verbatim in every bridged call; same 3 wrong answers |

Unit tests (`tests/test_opencode_prompt.py`) drive `execute()` with faked sandbox plumbing and assert the exact `exec_remote` call: the prompt (embedded quotes, blank lines, no trailing newline) is passed verbatim as `input`, is absent from argv, nothing but `opencode.json` is written to the sandbox, and the system-prompt prepend and `--continue` paths go through the same channel. `make check` clean; fast suite: 97 passed, 1 pre-existing failure unrelated to this change (`test_bundled_catalog_tracks_live_latest`: the live Codex catalog's latest moved to `gpt-6-astra`, bundled fallback still says `gpt-5.6-sol`).

### Agent review
- Reviewer: Claude Fable 5.1 via a fresh-context general-purpose subagent (same model as author), 1 pass
- Findings: 4 on the initial (staged-file) version — 2 fixed, 2 dismissed as nits. The automated PR review then flagged the leftover prompt file; that and the reviewer's file-related findings are moot after switching to `exec_remote(input=...)`, which removes the file and the bash wrapper entirely.
- Verification of the final version: 2-sample stand-in 2/2, and the 10 GAIA samples re-run (see table; final row added after the switch).
- Docs-only follow-up 28fb331 (opencode version floor, from ransomr's review note) authored by Claude Fable 5.1 as a CASE pr-developer worker; no separate review pass on it.

### DIFF
--- a/src/inspect_swe/_opencode/opencode.py
+++ b/src/inspect_swe/_opencode/opencode.py
@@ -1,6 +1,5 @@
 import json
 import shlex
-import uuid
 from pathlib import Path
 from textwrap import dedent
 from typing import Any, Literal, Sequence
@@ -266,13 +265,6 @@ def opencode(
                     state=state,
                 )
             else:
-                # per-invocation prompt file (see opencode_stdin_prompt_cmd); unique
-                # so concurrent opencode agents sharing a sandbox can't clobber each
-                # other, in a user-owned dir so the wrapper can unlink it
-                prompt_dir = f"{sandbox_home}/.inspect_swe/opencode"
-                prompt_path = f"{prompt_dir}/prompt-{uuid.uuid4().hex}.txt"
-                await sbox.exec(["mkdir", "-p", prompt_dir], user=user)
-
                 debug_output: list[str] = []
                 agent_prompt = prompt
                 attempt_count = 0
@@ -284,10 +276,6 @@ def opencode(
                     # the inbound state already carries an assistant turn)
                     if has_assistant_response or attempt_count > 0:
                         agent_cmd.append("--continue")
-
-                    # deliver the prompt on stdin rather than as a positional
-                    # argument (see opencode_stdin_prompt_cmd)
-                    await sbox.write_file(prompt_path, agent_prompt)
 
                     # Retry-loop gate: fires ONLY when this loop is actually
                     # retrying (attempt_count > 0), so the cold-start
@@ -302,9 +290,24 @@ def opencode(
                             required=True,
                         )
 
+                    # Deliver the prompt on stdin rather than as a positional
+                    # argument. `opencode run` quote-wraps a positional message
+                    # that contains spaces and backslash-escapes the double
+                    # quotes inside it (packages/opencode/src/cli/cmd/run.ts),
+                    # so a prompt passed as an argument reaches the model -- and
+                    # crosses the agent bridge -- as `"..."` with `\"` inside
+                    # rather than as the task input. The bridge anchors
+                    # main-thread tracking on the task input, and for prompts
+                    # containing `"` that mismatch let opencode's session-title
+                    # generation call displace the agent's answer as the sample
+                    # output. Piped stdin is used verbatim (`resolveRunInput`),
+                    # and also sidesteps argv length limits for long prompts.
+                    # exec_remote closes stdin after writing `input`, giving
+                    # opencode the EOF it needs.
                     result = await sbox.exec_remote(
-                        cmd=opencode_stdin_prompt_cmd(agent_cmd, prompt_path),
+                        cmd=agent_cmd,
                         options=ExecRemoteAwaitableOptions(
+                            input=agent_prompt,
                             cwd=agent_cwd,
                             env=agent_env,
                             user=user,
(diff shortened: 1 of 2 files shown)


dragonstyle:
Addressed the review nit in 07cc8b9: the wrapper now unlinks the prompt file once the redirect holds it open (test asserts it is gone), and the prompt directory is only created on the non-centaur path.

dragonstyle:
Reworked in f4776a7: the prompt now goes through `exec_remote(input=...)` (stdin over the sandbox-tools RPC channel), so there is no staged file and no bash wrapper at all. The unit tests now drive `execute()` against faked sandbox plumbing and assert the exact `exec_remote` call. PR description updated.

ransomr (approved):
Reviewed with an eye to regressions. Verified: `input` on `exec_remote` exists since inspect_ai 0.3.180 (pin is 0.3.255) and closes stdin after writing, so opencode gets its EOF and the old `exec 0</dev/null` wrapper is no longer needed (it would discard the prompt now). For opencode >= 1.15.0 `resolveRunInput` uses piped stdin verbatim when there is no positional message, so the task input crosses the bridge as an exact match; the title call always leads with "Generate a title for this conversation" and never anchors, so the real thread wins in either order.

One non-blocking note: opencode 1.2.x-1.14.x quote-wrap the positional but still do `message += "\n" + stdin`, so with this change the model sees a leading newline before the prompt. Prompts of 20+ chars still anchor by containment and beat the title call; quote-free prompts under 20 chars drop from the quoted grade to no match and the title call wins on length. Only reachable with an old opencode pre-installed in the image (`version="auto"` prefers the sandbox copy). Worth a docs line that stdin delivery needs opencode 1.15+, or a whitespace-tolerant fingerprint on the inspect_ai side.

dragonstyle:
Pinned the floor: `resolveRunInput` (piped stdin used verbatim when there is no positional) first shipped in opencode v1.14.42 (anomalyco/opencode#23557, 2026-05-08), not 1.15.0. Everything from 1.0.x through 1.14.41 does `message += "\n" + stdin`. 28fb331 documents that on the docs installation page, in the `version` docstring, and in the delivery comment. No runtime guard: it would cost an `opencode --version` exec per sample on the auto path, and falling back to argv for old versions would bring back the quote-escape mismatch for prompts containing `"`.

On the inspect_ai side I'd add a whitespace-tolerant exact match to `_position_descent` in #5284 (`stripped == initial_text` grades `EXACT`, right after the fingerprint check), which covers the old-opencode leading newline without touching the fingerprint hashing.

PR #152, opened 2026-09-09
feat: add Pi coding agent support

duarteocarmo:
- goal here was to add [pi](https://pi.dev/) as a harness to inspect_swe
- I think I followed all the practices in the repo - similar to other harnesses
- Pi is a bit specific (--yolo by default, etc) - but I think should fit nicely! 
- let me know what you think!


duarteocarmo:
@dragonstyle  - any interest in this? 

Thanks!

PR #164, opened 2026-09-17
fix: preserve Claude Code failure diagnostics without debug logging

timothee-chauvin:
Claude Code can emit a structured result/error record and then exit unsuccessfully while stderr contains only an unrelated warning. With debug disabled, the wrapper currently drops that structured evidence and raises with stderr alone.

Preserve a bounded summary of the latest invocation in a typed Inspect store and append it to nonzero-exit errors. The summary retains allowlisted result/error categories, validated timing/count metadata, malformed-record counts, and explicit missing/unclassified/truncated indicators. It excludes arbitrary error/result prose, prompts, tool arguments, session identifiers and unknown fields. Existing stderr handling remains unchanged.

Reset the summary before readiness and launch operations so failed relaunches cannot inherit an earlier invocation's evidence. Record it incrementally so received evidence survives stream failures. This closes a diagnostic gap; it does not diagnose the underlying crashes or change success, refusal, retry limits, warning handling or cancellation.

Validation: 383 tests passed, 64 skipped; repository-wide Ruff lint and formatting, strict mypy (131 files), and `git diff --check` passed. New coverage includes debug-disabled failures, missing/success/error results, privacy and bounds, malformed records, retry resets and limits, failed relaunches, stream failures and cancellation.

### Agent review

- Reviewer: independent Codex subagent in a fresh context; exact model identifier was not exposed in the review handoff. No different-model review is claimed.
- One initial review and one focused verification pass. The review found one P2 issue: resetting after subprocess launch retained stale diagnostics when a retry failed to launch. A regression reproduced it; moving reset before readiness/launch fixed it.
- Independent follow-up: 2 passed, 9 deselected; no new blockers. The author reran the full checks after the fix.


timothee-chauvin:
rogue agent, sorry

